PR #5, opened 2013-10-03
Add a encryption feature

bigeye:
When a user save journal file which has 'org-crypt-tag-matcher',
org-crypt will encrypt this entry. With this tag, Try to
encrypt/decrypt their journal whenever a user saves/opens their
journals.

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -77,6 +77,11 @@ string if you want to disable timestamps."
   "String that is put before every time entry in a journal file.
   By default, this is an org-mode sub-heading."
   :type 'string :group 'org-journal)
+
+(require 'org-crypt nil 'noerror)
+(when (fboundp 'org-crypt-use-before-save-magic)
+  (org-crypt-use-before-save-magic))
+
 (defcustom org-journal-enable-encryption nil
   "If non-nil, New journal entries will have a
 `org-crypt-tag-matcher' tag for encrypting. Whenever a user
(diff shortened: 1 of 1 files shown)


k0377:
Any updates on this? I would really like to see an encryption feature in my personal diary!

bastibe:
I'm afraid I currently don't have time to look into this. If anyone else would want to do this, I'd be grateful for it, though.

punchagan:
I [rebased](https://github.com/punchagan/org-journal/tree/org-crypt) this to work with the latest master.

punchagan:
I have added a customizable variable to determine when to do the encryption. Encryption is not setup automatically, now and needs to be done manually.

punchagan:
I've pushed a commit to make this happen automatically in `org-journal`. All one would need to do is to enable encryption by setting `org-journal-enable-encryption`.

bastibe:
Thank you for your work on this!

I don't quite like how this asks for the passphrase on every encrypt, but I guess this is in org-crypt's hand now. Anyway, I also undid the changes you did to the way files are opened/viewed--they looked good at first glance, but actually didn't work reliably. For more detail, see 1dd27c7005f8372f92c9571726a2793ba3c407f7.

punchagan:
I haven't yet tried this, but it looks like it should be possible to setup `gpg-agent` to cache pass-phrases.

Issue #27, opened 2015-02-26
Journal-wide search facilities


vkazanov:
At some point I realized that basic calendar intergration is not enough for the journal - manually searching dozens of files just does not scale very well. Agenda-based search does not scale to dozens of files too.

One option is to use generic Emacs tools, but this does not allow for smart queries such as "journal entries with the tag 'tag' during the previous week".

So, I am working on specialized search system for the org-journal. Currently my code is mostly tag-based (queries like "find entries with the following tags" are possible).

I am considering other, more powerful, means of searching (such as "find entries with a given text over a given period").

The question is: what do you, the creator of the package, use for searching and categorizing journal entries? This would help to adapt my code to your needs too.

bastibe:
That's a very interesting idea! I would *love* to be able to search a certain time-slice in the journal! Having a keyboard shortcut in the calendar that searches the current week or month would be genius!

At the moment, I am mostly using agenda searches, and relatively quickly-changing tags. For longer projects, I tend to move my development notes to a separate org file because agenda searches just don't work any more. But agenda searches really don't scale very well, and always open *all* journal files. It's not a perfect solution.

vkazanov:
I use the journal to write notes about things I did, movies I watched, code I wrote, daytime work notes, etc. Sometimes it is nice to get, say, a list of movie-related journal entries I watched last week... 

I guess that plain text search is okay here, as for tag searches I can just search for something like ":movie:".

So... How about the following: I will add a function doing a period-based search, with a few calendar-specific bindings doing month/week/etc searches? The interface will probably be a bit spartan, as I couldn't make up my mind about a proper way to display search results, it's just insert-text-button for now.

bastibe:
Very cool! I like this!

A simple interface is perfectly all right. If we later find a way to improve it, we can still add that later.

PR #28, opened 2015-02-27
#27 Basic journal search facility

vkazanov:
Here goes.

The main entry point is the `org-journal-search` function, which asks for a search string and two dates (start end). Dates are read using the `org-read-date` function, which allows a rather free-form date entry (like `-7d` or `-1m`) plus a calendar-based date selection.

The results are just a read-only buffer with clickable (mouse/keyboard) links. I am thinking about implementing a nicer mode for the result buffer, but for now this should do.

I also played with a tag-based search, which collects tags from all the files and builds an index... Don't know if it's worth it.

Please, give it a try and let me know.

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -381,6 +381,101 @@ If the date is not today, it won't be given a time."
     (org-journal-mark-entries)
     (calendar-exit)))
 
+;;; Journal search facilities
+;;
+
+(defun org-journal-search (str)
+  "Search for a string in the journal within a given interval.
+See `org-read-date` for information on ways to specify dates."
+  (interactive (list (read-string "Enter a string to search for: " nil 'org-journal-search-history)))
+  (let ((org-read-date-prefer-future nil)
+        (start (org-read-date nil t nil "Enter a period start"))
+        (end (org-read-date nil t nil "Enter a period end")))
+    (when (time-less-p end start)
+      (error "Period end cannot be before the start"))
+    (when (time-less-p (current-time) start)
+      (error "Period start cannot be in the future"))
+    (org-journal-search-by-string str start end)))
+(defvar org-journal-search-history nil)
+
+(defun org-journal-search-by-string (str &optional period-start period-end)
+  "Search for a string within a given interval"
+  (let* ((files (org-journal-search-build-file-list period-start period-end))
+         (results (org-journal-search-do-search str files)))
+    (with-current-buffer-window
+     "*Org-journal search*" nil nil
+     (princ (concat "Search results for \"" str "\": \n\n"))
+     (org-journal-search-print-results results))))
+
+(defun org-journal-search-build-file-list (&optional period-start period-end)
+  "Build a list of journal files within a given time interval"
+  (let ((files (directory-files org-journal-dir t
+                                org-journal-file-pattern))
+        result)
+    (dolist (file files)
+      (let ((filetime (org-journal-calendar-date->time
+                       (org-journal-file-name->calendar-date
+                        (file-name-base file)))))
+        (cond ((not (and period-start period-end))
+               (push file result))
+
+              ((and period-start period-end
+                    (time-less-p period-start filetime)
+                    (time-less-p filetime period-end))
+               (push file result))
+
+              ((and period-start
+                    (not period-end)
+                    (time-less-p period-start filetime))
+               (push file result))
+
+              ((and period-end
+                    (not period-start)
+                    (time-less-p filetime period-end))
+               (push file result)))))
+    result))
+
+(defun org-journal-search-do-search (str files)
+  "Search for a string within a list of files, return match pairs (PATH . LINENUM)"
+  (let (results)
+    (dolist (fname files)
+      (with-temp-buffer
+        (insert-file-contents fname)
+        (while (search-forward str nil t)
+          (let* ((fullstr (buffer-substring-no-properties
+                           (line-beginning-position)
+                           (line-end-position)))
+                 (res (list fname (line-number-at-pos) fullstr)))
+            (push res results)))))
+    results))
+
+(defun org-journal-search-print-results (results)
+  "Print search results using text buttons"
+  (dolist (res results)
+    (let* ((fname (nth 0 res))
+           (lnum (nth 1 res))
+           (fullstr (nth 2 res))
+           (time (org-journal-calendar-date->time
+                  (org-journal-file-name->calendar-date
+                   (file-name-base fname))))
+           (label (format-time-string org-journal-date-format time)))
+      (insert-text-button label
+                          'action 'search-journal-follow-link-action
+                          'org-journal-link (cons fname lnum))
+      (princ "\t")
+      (princ fullstr)
+      (princ "\n"))))
+
+(defun org-journal-search-follow-link-action (button)
+  "Follow the link using info saved in button properties"
+  (let* ((target (button-get button 'org-journal-link))
+         (fname (car target))
+         (lnum (cdr target)))
+    (find-file-other-window  fname)
+    (goto-char (point-min))
+    (forward-line (1- lnum))))
+
+
 (provide 'org-journal)
 
 ;;; org-journal.el ends here
--- a/org-journal.el
+++ b/org-journal.el
@@ -384,22 +384,68 @@ If the date is not today, it won't be given a time."
 ;;; Journal search facilities
 ;;
 
-(defun org-journal-search (str)
+(defun org-journal-search (str &optional period-name)
   "Search for a string in the journal within a given interval.
 See `org-read-date` for information on ways to specify dates."
   (interactive (list (read-string "Enter a string to search for: " nil 'org-journal-search-history)))
-  (let ((org-read-date-prefer-future nil)
-        (start (org-read-date nil t nil "Enter a period start"))
-        (end (org-read-date nil t nil "Enter a period end")))
-    (when (time-less-p end start)
-      (error "Period end cannot be before the start"))
-    (when (time-less-p (current-time) start)
-      (error "Period start cannot be in the future"))
+  (let* ((period-pair (org-journal-read-period period-name))
+         (start (car period-pair))
+         (end (cdr period-pair)))
     (org-journal-search-by-string str start end)))
 (defvar org-journal-search-history nil)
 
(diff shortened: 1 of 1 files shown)


org-journal.el
```diff
@@ -381,6 +381,101 @@ If the date is not today, it won't be given a time."
...
+  (dolist (res results)
+    (let* ((fname (nth 0 res))
+           (lnum (nth 1 res))
+           (fullstr (nth 2 res))
+           (time (org-journal-calendar-date->time
+                  (org-journal-file-name->calendar-date
+                   (file-name-base fname))))
+           (label (format-time-string org-journal-date-format time)))
+      (insert-text-button label
+                          'action 'search-journal-follow-link-action
```
bastibe: This should probably be `org-journal-search-follow-link-action`, right?

vkazanov: Yes :-) Sorry for that, I guess I had the original prototype function in the global scope left.

bastibe: I'm still looking at it ;-)

vkazanov: All right. I'll be around :-)

org-journal.el
```diff
@@ -381,6 +381,101 @@ If the date is not today, it won't be given a time."
...
+      (princ "\n"))))
+
+(defun org-journal-search-follow-link-action (button)
+  "Follow the link using info saved in button properties"
+  (let* ((target (button-get button 'org-journal-link))
+         (fname (car target))
+         (lnum (cdr target)))
+    (find-file-other-window  fname)
+    (goto-char (point-min))
+    (forward-line (1- lnum))))
```
bastibe: Somehow, the character position is lost since `find-file-other-window` implicitly collapses all subtrees.

Using 

```lisp
(org-journal-read-or-display-entry
     (org-journal-calendar-date->time
      (org-journal-file-name->calendar-date (file-name-base fname))))
```

instead fixes the issue.

org-journal.el
```diff
@@ -381,6 +381,101 @@ If the date is not today, it won't be given a time."
...
+           (fullstr (nth 2 res))
+           (time (org-journal-calendar-date->time
+                  (org-journal-file-name->calendar-date
+                   (file-name-base fname))))
+           (label (format-time-string org-journal-date-format time)))
+      (insert-text-button label
+                          'action 'search-journal-follow-link-action
+                          'org-journal-link (cons fname lnum))
+      (princ "\t")
+      (princ fullstr)
```
bastibe: I kind of don't like how this prints all the `***` prefixes if the search matches a title. This can be changed by removing beginning-of-line asterisks using `(princ (replace-regexp-in-string "^\*+\s*" "" fullstr))`. This is really a matter of taste though, and leaving the stars in is entirely fine by me!

bastibe:
This is incredibly cool and useful! I love how agenda searches take several seconds on my machine and open *all* org-journal files ever to do the search, while your code runs the same search in no time and without cluttering up my buffer list. This is awesome!

org-journal.el
vkazanov: Hm. Missed that. Besides, it is reasonable to use the viewing instead of finding, as we sort of read the entry archive.

org-journal.el
vkazanov: Well... I wouldn't do that. What for? After all, a user should be able to distinguing between the name of the tree node and the text of the node itself - the prefixes help with that.

bastibe:
If you fix the two or three issues mentioned above, I'll gladly merge this. Here's a feature request though: I would love a set of key bindings that initiate a search from the calendar in the current week, month, or year. Do you want to implement these before we merge?

vkazanov:
No problem, I'll do it along with the minor issues mentioned.

vkazanov:
I think I am done. 

Now there are four search interface functions: `org-journal-search`, `org-journal-search-calendar-week`, `org-journal-search-calendar-month`, `org-journal-search-calendar-year`. The latter three can only be called from the calendar-mode as they use a week/month/year currently under the cursor at the moment.

I did not come up with keybindings for the functions. Honestly speaking, I do not usually do things like searching using keybindings - prefer calling such commands through `M-x`.

vkazanov:
Can you offer convenient personally for you bindings in the calendar? I'd just add those, as I have no preferences of my own.

bastibe:
I just played around with it some more. This is really awesome! Thank you!

These are the bindings I came up with:
```
(define-key calendar-mode-map (kbd "i j") 'org-journal-new-date-entry)
     (define-key calendar-mode-map (kbd "f f") 'org-journal-search-calendar-forever)
     (define-key calendar-mode-map (kbd "f w") 'org-journal-search-calendar-week)
     (define-key calendar-mode-map (kbd "f m") 'org-journal-search-calendar-month)
     (define-key calendar-mode-map (kbd "f y") 'org-journal-search-calendar-year)
```

org-journal.el
```diff
@@ -381,6 +381,173 @@ If the date is not today, it won't be given a time."
...
+  (interactive (list (read-string "Enter a string to search for: " nil 'org-journal-search-history)))
+  (org-journal-search str 'week))
+(defun org-journal-search-calendar-month (str)
+  "Search for a string within a current calendar-mode week entries"
+  (interactive (list (read-string "Enter a string to search for: " nil 'org-journal-search-history)))
+  (org-journal-search str 'month))
+(defun org-journal-search-calendar-year (str)
+  "Search for a string within a current calendar-mode week entries"
+  (interactive (list (read-string "Enter a string to search for: " nil 'org-journal-search-history)))
+  (org-journal-search str 'year))
```
bastibe: I would like to add

```
(defun org-journal-search-forever (str)
  "Search for a string within all entries"
  (interactive (list (read-string "Enter a string to search for: " nil 'org-journal-search-history)))
  (org-journal-search str 'forever))
```

org-journal.el
```diff
@@ -381,6 +381,173 @@ If the date is not today, it won't be given a time."
...
+  (cond
+   ;; no period-name? ask the user for input
+   ((not period-name)
+    (let* ((org-read-date-prefer-future nil)
+           (absolute-start (time-to-days (org-read-date nil t nil "Enter a period start")))
+           (absolute-end (time-to-days (org-read-date nil t nil "Enter a period end")))
+           (start (calendar-gregorian-from-absolute absolute-start))
+           (end (calendar-gregorian-from-absolute absolute-end)))
+      (cons start end)))
+
```
bastibe: I would like to add

```
;; eternity start/end
   ((and (eq period-name 'forever) (eq major-mode 'calendar-mode))
      (cons (list 0 0 0)
            (list 21 31 3000)))
```

org-journal.el
```diff
@@ -381,6 +381,173 @@ If the date is not today, it won't be given a time."
...
+         (end (org-journal-calendar-date->time (cdr period-pair))))
+    (org-journal-search-by-string str start end)))
+(defvar org-journal-search-history nil)
+
+(defun org-journal-search-calendar-week (str)
+  "Search for a string within a current calendar-mode week entries"
+  (interactive (list (read-string "Enter a string to search for: " nil 'org-journal-search-history)))
+  (org-journal-search str 'week))
+(defun org-journal-search-calendar-month (str)
+  "Search for a string within a current calendar-mode week entries"
```
bastibe: This docstring should say `month` instead of `week`

org-journal.el
```diff
@@ -381,6 +381,173 @@ If the date is not today, it won't be given a time."
...
+  (interactive (list (read-string "Enter a string to search for: " nil 'org-journal-search-history)))
+  (org-journal-search str 'month))
+(defun org-journal-search-calendar-year (str)
+  "Search for a string within a current calendar-mode week entries"
```
bastibe: This docstring should say `year` instead of `week`

org-journal.el
```diff
@@ -381,6 +381,173 @@ If the date is not today, it won't be given a time."
...
+           (label (format-time-string org-journal-date-format time))
+
+           (label-end (format-time-string org-journal-date-format period-start)))
+
+      (insert-text-button label
+                          'action 'org-journal-search-follow-link-action
+                          'org-journal-link (cons fname lnum))
+      (princ "\t")
+      (princ fullstr)
+      (princ "\n"))))
```
bastibe: I'd like to add a few key bindings for this buffer:

```
(local-set-key (kbd "q") 'kill-this-buffer)
(local-set-key (kbd "<tab>") 'forward-button)
(local-set-key (kbd "<backtab>") 'forward-button)
```

bastibe:
My comments don't seem to work correctly. I'm still trying to figure this out.

org-journal.el
```diff
@@ -381,6 +381,173 @@ If the date is not today, it won't be given a time."
...
     (calendar-exit)))
 
+;;; Journal search facilities
+;;
+
+(defun org-journal-search (str &optional period-name)
+  "Search for a string in the journal within a given interval.
+See `org-read-date` for information on ways to specify dates."
+  (interactive (list (read-string "Enter a string to search for: " nil 'org-journal-search-history)))
+  (let* ((period-pair (org-journal-read-period period-name))
```
bastibe: At this point, I think that a prefix argument that switches from a bounded search to a search across all files is really useful:

```
(let* ((period-pair (org-journal-read-period (if current-prefix-arg 'forever period-name)))
...
```

bastibe:
After playing around with this a bit, I found that I often wanted to search all journal entries without explicitly mentioning a date range. I added comments on how to do this.

See https://www.dropbox.com/s/uoeiuii0s32abkw/patch for a patch containing the mentioned changes.

vkazanov:
Great, I'll add the changes and see if there any problems.

vkazanov:
Works for me now. I have a few more minor improvement ideas, but they'l wait - got to test the current version first.

bastibe:
Great! I love it! Just say the word and I'll merge it.

vkazanov:
The word? :-D

vkazanov:
I believe it is more or less usable now, so yes, go on :-)

bastibe:
Very cool! Thank you so much for this! After playing around with it in the last two days, I'm already wondering how I could use org-journal without it beforehand.

Should I upload this as a new release, or do you want to work some more on it?

vkazanov:
It's up to you, I guess. I will fix bugs, if I happen to find any, but the main mechanics is out there already.

PR #35, opened 2015-04-21
Two minor improvements for org-mode

vkazanov:
1. The first improvement is plain and simple: with a prefix arg `org-journal-new-entry` opens a current entry. If there's no entry - it opens an empty file, which is clearly wrong. The first commit fixes the problem by making the function throw an error in the situation.

2. The second improvement is a bit controversial. 

A have my journal files stored in a Dropbox directory (~/Dropbox/Documents/journal), which allows for sharing between my home and office machines. My ~/Documents directory is just a symlink to ~/Dropbox/Documents (on both office Windows and home Linux machines).

Now, if I set org-journal-dir to ~/Dropbox/Documents/journal, then (buffer-file-name) in any of the journal entry files does not match the auto-mode-alist regex (filename = "~/Documents/journal/...", regex = "~/Dropbox/Documents/.....").

If I set org-journal-dir to to ~/Documents/journal then org-journal-update-auto-mode-alist expands it to ~/Dropbox/Documents/journal anyway, because it uses `file-truename`, which expands the all the symlinks. So, I changed file-truename for `expand-file-name`, which works like charm.

And I could not come up with a situation where the change would hurt.

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -57,7 +57,7 @@
 (defun org-journal-update-auto-mode-alist ()
   "Update auto-mode-alist to open journal files in
   org-journal-mode"
-  (let ((name (concat (file-truename org-journal-dir)
+  (let ((name (concat (expand-file-name org-journal-dir)
                       (substring org-journal-file-pattern 1))))
     (add-to-list 'auto-mode-alist
                  (cons name 'org-journal-mode))))
--- a/org-journal.el
+++ b/org-journal.el
@@ -187,29 +187,32 @@ without adding an entry"
   (org-journal-dir-check-or-create)
   (let* ((entry-path (concat org-journal-dir
                              (format-time-string org-journal-file-format)))
-         (entry-path-exists-p (file-exists-p entry-path)))
-    ;; a prefix arg is given and there's no current journal entry? error, as
-    ;; there's no current entry to open
-    (when (and prefix
-               (not entry-path-exists-p))
-      (error "There's no current entry to open."))
-    (find-file entry-path))
-  (goto-char (point-max))
-  (let ((unsaved (buffer-modified-p)))
-    ;; skip entry adding if a universal prefix is given
-    (unless prefix
-      (if (equal (point-max) 1)
-          (insert org-journal-date-prefix
-                  (format-time-string org-journal-date-format)))
-      (unless (eq (current-column) 0) (insert "\n"))
-      (insert "\n" org-journal-time-prefix
-              (format-time-string org-journal-time-format)))
-    (org-journal-mode)
-    (hide-sublevels 2)
-    ;; open the most recent entry
-    (when prefix
-      (show-entry))
-    (set-buffer-modified-p unsaved)))
+         (entry-path-exists-p (file-exists-p entry-path))
+         (should-add-entry-p (not prefix)))
+    (find-file entry-path)
+    (goto-char (point-max))
+    (let ((unsaved (buffer-modified-p)))
+
+      ;; empty file? Add a date timestamp
+      (when (equal (point-max) 1)
+        (insert org-journal-date-prefix
+                (format-time-string org-journal-date-format)))
+
+      ;; skip entry adding if a prefix is given
+      (when should-add-entry-p
+        (unless (eq (current-column) 0) (insert "\n"))
+        (insert "\n" org-journal-time-prefix
+                (format-time-string org-journal-time-format)))
+
+      ;; switch to the outline, hide subtrees
+      (org-journal-mode)
+      (hide-sublevels 2)
+
+      ;; open the recent entry when the prefix is given
+      (unless should-add-entry-p
+        (show-entry))
+
+      (set-buffer-modified-p unsaved))))
 
 (defun org-journal-calendar-date->time (calendar-date)
   "Convert a date as returned from the calendar to a time"
--- a/org-journal.el
+++ b/org-journal.el
@@ -185,8 +185,15 @@ Giving the command a prefix arg will just open a today's file,
 without adding an entry"
   (interactive "P")
   (org-journal-dir-check-or-create)
-  (find-file (concat org-journal-dir
-                     (format-time-string org-journal-file-format)))
+  (let* ((entry-path (concat org-journal-dir
+                             (format-time-string org-journal-file-format)))
+         (entry-path-exists-p (file-exists-p entry-path)))
+    ;; a prefix arg is given and there's no current journal entry? error, as
+    ;; there's no current entry to open
+    (when (and prefix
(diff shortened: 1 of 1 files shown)


bastibe:
Thank you for this!

Could you give me an example where the first change improves one of your workflows? When is opening an empty file the wrong answer? Personally, I never run into this--I usually use org-journal-new-file with prefix as a "View today's entry".

Your second change is very cool! I don't use a custom org-journal-dir, so I don't have this problem, but expand-file-name is clearly the correct choice.

vkazanov:
Well... I does not improve on anything workflow-related, just makes sure the entry is there. Every now and then I do `C-u C-c C-j` just to see an empty non-existant file. 

Don't know. Maybe it's more reasonable to ask the user if org-journal should create a new entry in this case?  Instead of throwing an error?

d12frosted:
Wanted to comment first paragraph. I think that `C-c C-j` creating new file is really right way to do, because it stands for 'create new entry'. And it should just create new file when it doesn't exist.

On other hand, I agree that if user just wants to visit journal file - it should not create it.

While I was doing some updates to `org-journal` I asked myself why 'visit todays journal file' and 'create new entry' are actually handled by the same function. And I actually asked myself, why is it creating new file when `prefix` is passed. 

I think that right way for that is creating two separate functions. One for creating new entry that will create new entry no matter what (and currently it does it perfectly). Another one for visiting files (in read-only mode?).

How about that? In case you don't have time for that I can handle it :D

bastibe:
Yes, that's kind of what I was thinking, too. And actually, that read-only function already exists (`org-journal-read-or-display-entry`). It is bound to `j` in the calendar, and opens the file in `view-mode`. If the file doesn't exist, an error message is thrown.

Maybe we can settle for a compromize: How about we rebind `C-u C-c C-j` to `org-journal-read-or-display-entry`? This would do one of three things:
- If no entry exists, it shows an error message.
- If an entry exists and the file is already opened, it is shown as-is.
- if an entry exists and is not already opened, it is opened in view-mode.

Does that sound reasonable?

d12frosted:
Ah, I played around with my suggestion a bit. And it turns out to be not as effective as I thought. I mean, separating this into two functions. Because it breaks a bit current org-journal usage: `C-c C-j` to create new entry and `C-u C-c C-j` to visit it. So I am suggesting doing something like:

```lisp
(defun org-journal-new-enty (prefix)
  (interactive "P")

**Comment by vkazanov** [comment]
Bastibe,

wait! The difference between `C-u C-c C-j` and `C-c C-j` is not in "read" or "display". It is about creating a new entry, or just visiting the recent one.

The whole point is that I edit the last entry a lot (I usually keep today's TODO list in it). Switching this `C-u C-c C-j` to the `read-or-display` would completely break this workflow. Please, don't do this :-)

The only thing I wanted to do is make sure today's entry file really exists when doing `C-u C-c C-j` :-)

**Comment by vkazanov** [comment]
d12frosted,

how is your suggestion different from the current definition?

```
(defun org-journal-new-entry (prefix)
  "Open today's journal file and start a new entry.
Giving the command a prefix arg will just open a today's file,
without adding an entry"
  (interactive "P")
```

**Comment by d12frosted** [comment]
@vkazanov 

Sorry about confusion I made.

As I said in my last comment

> `C-c C-j` to create new entry and `C-u C-c C-j`

That's why I am proposing to

```lisp
(defun org-journal-new-enty (prefix)
  (interactive "P") 
  (if prefix (visit) (create-new)))
```

But I find @bastibe 's suggestion better:

> Maybe we can settle for a compromise: How about we rebind `C-u C-c C-j` to `org-journal-read-or-display-entry`?

vkazanov:
@d12frosted 

This would completely break my workflow, see my comment above.

bastibe:
It seems that there can't be a compromise, then. 

The question remains what `C-u M-x org-journal-new-entry` should do when no current file exists. From the function name, I would assume that it creates a new, empty, file. Maybe it could create a today's file, but not add any time-subheadings? That would also make sense to me.

vkazanov:
@bastibe 

maybe it is reasonable to ask the user if he wants to add a new entry? I.e., say something like "There are no entries today. Create an entry?". If the user agrees - create a entry, disagrees - abort.

My main point here is that an *empty file in the org-journal system is an invalid state*, which should be avoided, and this is why I offered this change.

bastibe:
I agree with you on that: An empty file is not useful.

How about adding the *day*-header, but not the *time*-header, if the prefix is given?

vkazanov:
Well... How useful is that? :-)

I can see the following cases here:

1) If the prefix was specified by mistake -> a user wants to create an entry, and asking him is reasonable.

2) If the prefix was specified on purpose -> a user wants to see the current entry, which is absent. So he can make a decision to either abort or create the entry.

A similar thing happens when a user tries to visit a non-existing journal directory, so this is a consistent behaviour.

Well... The decision is yours, of course.

bastibe:
> Well... How useful is that? :-) A daily timestamp? :-)

I mean the first line in every file. The one that says, for example, `* Friday, 04/24/15`. I sometimes want to open today's file, but want to create a new entry that is not a time-entry. This would help with that.

vkazanov:
...but this also means that org-journal will have to open a fresh file with no entries, and add some text into it, which still is a bit incomplete. Later, the user will have to either add a journal entry by hand or call another command. 

To my taste and workflow having a warning and a possibility to resolve the problem at once ("No entry -> created one?") feels like a better solution.

Anyway, this is becoming too lengthy for "two minor issues". :-) You should probably just make up your mind -> I can always just have a little customization of my own.

bastibe:
Alright, I prefer the date-but-no-time-stamp version, then. Would you like to implement that, or should I do it?

vkazanov:
I'll do it, no problem. Writing Lisp is always a pleasure :-)

You don't mind if I add a customization facility? Something like an ability to choose an empty file action? The default would be to create the file, like you want, and an option - to ask if a new entry is wanted.

bastibe:
Great idea! This way, we can both have our way!

vkazanov:
So, basically now `org-journal-new-entry` does what you wanted it to do: without a prefix it inserts a new journal entry, with a prefix and no today's entries it opens a new file with a date header inserted.

I did not want to overcomplicate the code with the customization option - it just doesn't feel natural here.

bastibe:
Thank you very much! If you find a nice solution for implementing your workflow, please let me know and I'll include it in the readme. It seems that an [Advice](https://www.gnu.org/software/emacs/manual/html_node/elisp/Advising-Functions.html) should be enough.

My computer broke the other day, and it will take me a few days until I can test, merge, and publish this PR.

PR #36, opened 2015-04-23
improve auto hiding of time entries

d12frosted:
* define 'org-journal-do-hide-entries-on-new variable that is used to
  decide either to hide time entries or not when creating new one using
  'org-journal-new-entry function
* don't hide second level when creating new entry, but instead try to
  figure out the level of time entries from 'org-journal-time-prefix
  variable

---

Some explanations. 

**Visibility**

I thought that it might be good idea to let user to setup org-journal in a way that respects visibility of file when creating new entry. Personally, I like seeing previous notes, because I usually run in full-screen mode. 

**Figure level**

When I was adding support of visibility respect I noticed that function that creates new entry actually hides second level no matter what is set to `org-journal-time-prefix`. So like if even users uses `"* "` as `org-journal-time-prefix` it will hide second level instead of first (I think that this is what was intended, right?). So I updated it by creating new function that tries to figure out level from `org-journal-time-prefix` value. When it fails to do so - it just return `nil`, so nothing hides. 

---

My elisp-fu is no very good, so probably it all might be implemented in better way. In that case - consider my pull request as open feature request :D Also I apologise if names I used for new variable and new functions are bad - feel free to changes them (as well as description to them). 

Cheers, Boris.

P. S. Thank you for you package - it helps a lot in daily routine!

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -193,24 +193,28 @@ without adding an entry"
   (find-file (concat org-journal-dir
                      (format-time-string org-journal-file-format)))
   (goto-char (point-max))
-  (let ((unsaved (buffer-modified-p)))
-    ;; skip entry adding if a universal prefix is given
-    (unless prefix
-      (if (equal (point-max) 1)
-          (insert org-journal-date-prefix
-                  (format-time-string org-journal-date-format)))
+  (unless prefix
+    (let ((unsaved (buffer-modified-p))
+          (date-string (concat org-journal-date-prefix
+                               (format-time-string org-journal-date-format)))
+          (level (org-journal-figure-time-entry-level)))
+      (unless (buffer-contains-substring? date-string)
+        (insert date-string))
       (unless (eq (current-column) 0) (insert "\n"))
       (insert "\n" org-journal-time-prefix
-              (format-time-string org-journal-time-format)))
-    (org-journal-mode)
-    (when org-journal-do-hide-entries-on-new
-      (let ((level (org-journal-figure-time-entry-level)))
-        (when level
-          (hide-sublevels level)
-          ;; open the last entry
-          (when prefix
-            (show-entry)))))
-    (set-buffer-modified-p unsaved)))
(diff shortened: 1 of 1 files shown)


bastibe:
This is a very good idea!

The auto-hiding has in fact bothered me as well, although never enough to actually make it more reasonable. If it's OK with you, all refactor this a bit before uploading.

d12frosted:
I am glad that you like the idea. And as I already told - feel free to make whatever you think is needed before uploading :smile:

d12frosted:
Ok. Sorry I did one more change.

You see, I have a hook for `org-mode` that inserts for me template to empty files (actually, to files that doesn't contain this template). So whenever I create first entry in the day it actually does this:

1. `org-mode` creates template for the file 
2. `org-journal-new-entry` creates new file and sees that the position is not at `1` and doesn't insert `date-prefix` + date. 

So I did a dirty hack. Instead of checking position it checks if file contains that string (`date-prefix` + date). And if it's not - inserts. 

And again, probably thats not the way it should be implemented, but at least I described problem, provided sample solution, so now feel free to do whatever you think is needed :D 

P. S. oh and actually I did move some code around.

bastibe:
I hope you agree with the changes I made to make this conform with #35. I *think* that this should only insert a date entry if the file is empty, so your template workflow should work. Please let me know if this is what you intended.

d12frosted:
Thank you very much. And sure I don't mind of any changes you made. I'll test it as soon as possible and will let you know. 

```lisp
;; empty file? Add a date timestamp
      (when (equal (point-max) 1)
        (insert org-journal-date-prefix
                (format-time-string org-journal-date-format)))
```

This is the only thing that might bother me. I mean, it adds date header only when file is empty. So it might not work with templates. But it's OK. I honestly working on improving my current flow, so that's really not a problem.

Thanks again! :D

d12frosted:
Yeah, changed a bit my auto-insertion function and now it works like a charm. Thanks a lot!

PR #42, opened 2015-06-21
Use expand-file-name instead of concat.

punchagan:
org-journal-dir may not end with a / and using concat doesn't work
correctly, in that case.


bastibe:
This is much better, thanks!

PR #45, opened 2015-06-30
Encrypt entry created from calendar view.

punchagan:
Also de-duplicate code between `org-journal-new-entry` and
`org-journal-new-date-entry`.


bastibe:
Thank you for this contribution. I always wanted to de-duplicate these code paths. I re-organized your changes a bit, so we don't need that new function any longer.

PR #55, opened 2015-12-12
Add highlight to org-journal-search-print-results

aJchemist:


bastibe:
Thank you for this idea! This is pretty cool!

This requires highlight.el to be installed, right? I would prefer to not include this dependency. Do you know if there is a way to achieve this highlight without using highlight.el?

aJchemist:
@bastibe Ooops! I knew highlight.el is built-in emacs library. For now I have commited a workaround.

bastibe:
Very cool, this is a neat solution. Do you want to look for a way to do this without using highlight.el, or rather push this as a new version as soon as possible? 

It is probably relatively easy to implement this highlight using overlays, if you want to try your hand at that.

aJchemist:
@bastibe ping

vkazanov:
@aJchemist 

Just  my two cents. 

1. Great idea! I should have done this from start (I implemented the search).

2. It is generally better to use text properties for highlighting anything as overlays become slow really, REALLY easy. 

There some single linked lists behind the overlay mechanism, which are always linear.  I even discussed the problem on emacs-devel mailing list and will probably switch to a tree some day sunny day, but for now it's only useful for 5-10 overlays/buffer.

aJchemist:
@vkazanov Nice point :+1: thx for telling your clean solution.

bastibe:
Sorry, I was on vacation and did not check my emails. I like this a lot, though!

bastibe:
I am so sorry for not uploading this to melpa/marmalade earlier! I completely forgot about this.

I am uploading a new version right now.

aJchemist:
@bastibe nvm, luckily I'm not a user of emacs package system; melpa/marmalade or others.

Anyway. cool!

PR #68, opened 2016-10-22
Defcustom org-journal-search-results-order-by

aJchemist:
This is a light suggestion PR


bastibe:
I like it! Thank you very much!

PR #74, opened 2017-02-02
Add Spacemacs layer guide to README

borgnix:
Hi. I really enjoy the journaling experience with `org-journal`, so I made a Spacemacs layer to provide Spacemacs-style key binding wrapping. I would like to add a spacemacs layer guide to README in case that there are other Spacemacs users who need it.
Again, thanks for your work on `org-journal`.

### DIFF
--- a/README.org
+++ b/README.org
@@ -61,6 +61,11 @@ should be trivial:
 
 Then add =(require 'org-journal)= to your =.emacs=.
 
+*** Spacemacs Layer for org-journal
+
+To use =org-journal= on Spacemacs with Spacemacs-style key bindings, please check
+[[https://github.com/borgnix/spacemacs-journal][journal]] layer.
+
 ** Quickstart
 
 Doing =C-x org-journal-new-entry= (or =C-c C-j=) will immediately create a journal
--- a/README.org
+++ b/README.org
@@ -63,8 +63,12 @@ Then add =(require 'org-journal)= to your =.emacs=.
 
 *** Spacemacs Layer for org-journal
 
-To use =org-journal= on Spacemacs with Spacemacs-style key bindings, please check
-[[https://github.com/borgnix/spacemacs-journal][journal]] layer.
+To use =org-journal= on Spacemacs, you could :
+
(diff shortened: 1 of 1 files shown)


bastibe:
Very nice! Could this spacemacs layer be included directly in org-journal without breaking non-spacemacsen's configs?

borgnix:
This layer is specifically for Spacemacs users. It functions well with spacemacs layer mechanism, provides out of box `org-journal` experience for spacemacs users, but would be useless without spacemacs. It will not affect non-spacemacs' configs.

bastibe:
Could we include this layer into org-journal itself, or do you want to keep it in your own repository?

borgnix:
Tthe layer is a wrapper for org-journal. It's not an enhance to `org-journal` itself. It could be integrated into org-journal repo, but it would be totally useless for non-spacemacs users, and not very convenient for spacemacs users to install it(by cloning the whole repo into `.emacs.d/private` and let melpa manage `org-journal` package). so I think it would be better to keep it separate.

bastibe:
OK, thank you for the clarification.

borgnix:
Sorry, I'm not very good at writing README. Would it be okay if i change it to:

To use org-journal on spacemacs, you could :

1. `git clone https://github.com/borgnix/spacemacs-journal.git ~/.emacs.d/private/journal`
2. add it to your ~/.spacemacs. You will need to add `journal` to the existing dotspacemacs-configuration-layers list in this file.

Manuel of `journal` layer could be found at https://github.com/borgnix/spacemacs-journal

bastibe:
Of course, that would be very good. 

- Please set all variable names and paths as code
- Please change the last sentence to "The manual of the `journal` layer can be found at ..."

borgnix:
Okay, thanks for the review.

bastibe:
Thank you very much!

vkazanov:
@bastibe , how about moving Spacemacs-related comment into FAQ? Or to the bottom of the README? It feels a bit out of place near Synopsys. :-)

bastibe:
Good thinking. Will do.

bastibe:
done.

vkazanov:
Cool! :-) FAQ doesn't sound so negative now :-)

PR #75, opened 2017-02-21
A little hook called whenever an entry is created

vkazanov:
I want to force myself into always adding tags to my journal entries. So a little hook needed that would allow to run some custom code upon entry creation.

Also, some Elisp style fixes.


bastibe:
I like it! And I particularly like your elisp style fixes! My day-to-day work does not include much lisp, regrettably, and I am grateful for every opportunity for learning.

vkazanov:
Not too many people do have a chance to work with Lisps these days! Well, apart from those Clojure hipsters :-)

Issue #80, opened 2017-06-27
Evaluate org-journal-date-prefix


psamim:
Hi,

I wanted to be able to add `calendar-persian-date-string` at the top of my diary file. I thought I should set `org-journal-date-prefix`, but it expects a string.

Is there any way I could achieve this?

Thanks

bastibe:
The date entry is composed of the `org-journal-date-prefix`, and after that, the current date, as formatted by `org-journal-date-format`. `org-journal-date-prefix` is copied verbatim, and `org-journal-date-format` is passed to `format-time-string`. I would hope that `format-time-string` has some provisions for printing persion date strings.

psamim:
Unfortunately I could not find any.
I changed source code as follow for my self:
```emacs-lisp
      ;; empty file? Add a date timestamp
      (when new-file-p
        (insert org-journal-date-prefix
                (calendar-persian-date-string)
                " -- "
                (format-time-string org-journal-date-format time)))
```
It would be nice if we could have a general solution for evaluating some function at that point.

PR #82, opened 2017-06-30
Add org-journal-prefix-function

psamim:
This to fix https://github.com/bastibe/org-journal/issues/80.

Now user can set custom functions for prefix. E.g.:
```emacs-lisp
  (defun psamim-journal-prefix()
    (concat
     "* "
     (format-time-string "%B %e, %Y") "\n"
     (format-time-string "%A") "\n"
     (calendar-persian-date-string) "\n"
     (calendar-bahai-date-string) "\n\n"))

  (custom-set-variables
   '(org-journal-prefix-function 'psamim-journal-prefix))
```

### DIFF
--- a/README.org
+++ b/README.org
@@ -124,7 +124,8 @@ Customization options related to journal directory and files:
 Customization options related to the journal file contents:
 
 - =org-journal-date-format= - date format format =org-journal= uses when showing a
-  date within a journal and search results page.
+  date within a journal and search results page. If set to a function, it is evaluated 
+  and inserted.
 
 - =org-journal-date-prefix= - this string will prefix the date at the top of a journal
   file.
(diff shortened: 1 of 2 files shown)


bastibe:
This is cool! Thank you very much!

I think the function should accept `time` as an argument, though. Otherwise, it will do the wrong thing if invoked through the calendar on a day that is not today. In your example, you would have to pass `time` to both instances of `format-time-string` as a second argument.

Also, it might be more consistent to make this an alternative value for `org-journal-date-format` instead of its own variable. We could then check if the value is a function or a string, and either call it or pass it to `format-time-string`.

psamim:
I changed things as said. Please take a look.

bastibe:
Cool, I like it! Thank you very much!

PR #86, opened 2017-11-02
added autoload to org-journal-search

CeleritasCelery:
I have encountered an issue several times where I start Emacs and try to search for my journal for a keyword but `org-journal-search` is undefined. I have to open a journal before I can search.


bastibe:
Thank you! I only ever run `org-journal-search` from the calendar, which is why this never came up for me.

Issue #53, opened 2015-08-25
Can't use org-agenda before opening a journal file


Titan-C:
I followed the instructions to add all those calendar/journal files to org-agenda. But for some reason I need to have previously have opened one. If I call org-agenda as soon as I start emacs I get the message
```org-agenda-get-day-entries: Agenda file /home/oscar/Dropbox/org/journal/20150819 is not in `org-mode'```

bastibe:
Yes, `org-journal` is autoloaded only once it is interacted with. You could try `(add-hook 'org-mode-hook (lambda () (require 'org-journal)))` in your `init.el` and see if that fixes it.

bastibe:
Did this fix your problem?

Titan-C:
It didn't help or I didn't do it right at that time. But nowadays I have moved into a much more simplified workflow. I just use an org-mode capture template and put it all my journal notes in one same org file. So I have stopped using the features of this project, as my simple way is fine for now.
Thank you for your concern.

PR #88, opened 2017-11-08
fix org-journal-dir issue

CeleritasCelery:
Currently if `org-journal-dir` does not end in a directory separator (`/`) it will not add the journals to `auto-mode-alist`. There are other org directory variables (such as `org-directory`) that don't have the trailing directory separator, which makes this behavior unexpected. 

This fix will also correctly address #53, which isn't related to auto-loading.


bastibe:
Cool, thank you!

PR #97, opened 2018-03-28
Add example capture template to the README

### LINKED ISSUE #50: Integrate with org-capture
Is it possible to make an org-capture template that matches the settings of org-journal?

There seems to be some feature overlap between both packages, but I'm not enough of an elisper to tackle that…

Miciah:
Fixes #50.


bastibe:
Thank you very much! This question has come up several times, and now we have an answer!

Issue #61, opened 2016-01-23
Option to add entries past midnight to previous day with org-extend-today-until?


namdnguyen:
This option would be helpful for those who work late nights past midnight and want journal entries to be logged on the journal for the date before midnight. Org-mode already provides support for extending today, so maybe have an option to toggle whether to use that time or go with the default of midnight?

bastibe:
Interesting idea. I'll look into it, and see how difficult this would be to implement.

bastibe:
Sorry that this is taking a while, but my spare time is limited as of late.

leezu:
I'd also like to express "interest" in this features. (No worries / pressure though.)

nickanderson:
+1

bastibe:
Is anyone in here interested in trying to implement this? Pull requests would be very welcome!

TinaRussell:
I would really love this feature because the “Tina day” definitely ends somewhere past midnight.
What functions would need to be changed? I’ve long been tempted to take a swing at this, but I’m an Emacs/Elisp newb and I’d be afraid that changing one thing would collapse the whole org-journal edifice.

bastibe:
Wow, a lot of people seem to want this!

I implemented it in the latest version. Can you check that it works as intended? If you can confirm that it works, I'll push it to marmalade and melpa stable.

TinaRussell:
It’s having problems for me—it works as intended, except that entries past midnight (but before the org-extend-today-until hour) have a blank datestamp.
The problem seems to happen in here:

`      (when should-add-entry-p
        (unless (eq (current-column) 0) (insert "\n"))
        (let ((timestamp (if (= (time-to-days (current-time)) (time-to-days time))
                             (format-time-string org-journal-time-format)
                           "")))
          (insert "\n" org-journal-time-prefix timestamp))
        (run-hooks 'org-journal-after-entry-create-hook))`

The “if” statement returns nil, so the timestamp variable is set to "". This doesn’t seem to happen for the rest of the day, though.

bastibe:
Is this issue resolved?

bastibe:
Cool!

PR #110, opened 2018-09-01
Fix org-extend-today-until timestamp problems, add new option for post-midnight timestamps

TinaRussell:
This fixes issue #61 completely, by changing the “if” that decides whether or not an entry gets a timestamp to a “cond” that decides which kind of timestamp the entry should have (if any).

New local variables are added inside “org-journal-new-entry”:
“oetu-active-p”: Whether or not “org-extend-today-until” is being invoked—this required a big let statement around the entire function contents, and as a result the indentation is changed for the whole function.
“day-discrepancy”: The value of “(- (time-to-days (current-time)) (time-to-days time))”, established so that this check won’t need to be run for each cond clause, and it makes the code a little more readable.

New option “org-journal-time-format-post-midnight” allows the user to establish a different timestamp format for after midnight (when using “org-extend-today-until”), if desired.

…Not sure why the commits are credited to “Tina Russell and Tina Russell” in GitHub, with neither linking to my profile, weird. I’m, uh, new at this. But, I’ve tested the code thoroughly and it’s ready to go!

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -164,6 +164,11 @@ org-journal. Use org-journal-file-format instead.")
 string if you want to disable timestamps."
   :type 'string :group 'org-journal)
 
+(defcustom org-journal-time-format-post-midnight ""
+  "When non-blank, a separate time format string for after midnight,
+when the current time is before the hour set by `org-extend-today-until`."
+  :type 'string :group 'org-journal)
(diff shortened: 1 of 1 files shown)


bastibe:
Cool, thank you!

Issue #112, opened 2018-09-20
Automatic exit once entry is saved


dhruvparamhans:
Hello, 

Is there a way to automatically exit the current journal entry once I hit save? The kind of functionality I have in mind is similar to org-capture where you can exit with a simple key combination.

bastibe:
No, but you could simply rebind C-x C-s in org-journal-mode-map to do what you want.

dhruvparamhans:
Thanks for the suggestion. I cooked up something like this. I shall leave it here for others. 
```
(defun save-and-kill-org-journal ()
  "Saves the buffer and kills the window when in org-journal-mode"
  (interactive)
  (save-buffer)
  (kill-buffer-and-window))
(define-key org-journal-mode-map (kbd "C-x C-s") 'save-and-kill-org-journal)
 ```

By the way do you think it is worth adding this option by default in org-journal?

bastibe:
Cool! If you like, you could create a pull request and add this to the FAQ section of the readme.

bastibe:
I'm sorry I was imprecise in what I said: I meant adding an entry to the FAQ only. The code you wrote should go into the FAQ as well. This kind of customization is better left to the user, and does not need to be part of org-journal itself.

cslux:
@dhruvparamhans We added a section **Convenient =org-journal= Snippet Extensions** to the README with your snippet, and indicated that it comes from you.

Thank you!

PR #114, opened 2018-09-21
Update FAQ for keybind override

dhruvparamhans:
Save buffer is overwritten with save and exit buffer


bastibe:
This was added to the README in #112. Thank you, @dhruvparamhans for helping out!

PR #128, opened 2019-01-05
If encryption is enabled call `org-decrypt-entry` before searching fo…

cslux:
…r string


bastibe:
Thank you for your pull request. I don't personally use encryption.

How is performance when you do this? We intentionally avoided opening the files in org-mode, as that utterly destroys search performance. How does org-decrypt-entry deal with files that are not in org-mode?

cslux:
> We intentionally avoided opening the files in org-mode, as that utterly destroys search performance. How does org-decrypt-entry deal with files that are not in org-mode?

`org-decrypt-entry` keeps the buffer in fundamental mode, while org-journal-decrypt doesn't work when not in `org-mode` or `org-journal-mode`. `org-decrypt-entry` searches for the gpg header, and does some propertizing afterwards.  I have just assumed there is only one encrypted section, otherwise we would need in a loop.

Well, the decryption takes a while, but there is not much you can do about except putting all in one file (year journal...).

Without decryption the search takes 0.00357 seconds (14 Journal entries).
With decryption the search takes 2.2 seconds (14 Journal entries).
Based on this, searching through encrypted journal files is 600 times slower. But if the user wishes to use encryption, he must leave with that. This was the reason why I wished there was something like a year journal, etc., to speed this up.

cslux:
Right! 

On January 6, 2019 3:23:55 PM GMT+01:00, Bastian Bechtold <notifications@github.com> wrote:
>OK, cool. So it won't work for heave users, but at least the weekly or
>monthly search should work reasonably well. 

-- 
Christian Schwarzgruber
T: +43 699 194 166 34
E: c.schwarzgruber.cs@gmail.com

PR #129, opened 2019-01-05
Fix byte-compile warnings

cslux:


cslux:
`org-add-hook` looked like this before, so yes it is compatible with older Emacs versions.
``` emacs-lisp
(defun org-add-hook (hook function &optional append local)
  "Add-hook, compatible with both Emacsen."
  (if (and local (featurep 'xemacs))
      (add-local-hook hook function append)
    (add-hook hook function append local)))
```
And for `org-set-tags-to` I create an alias in case org version is smaller than 9.2, so yes this will also be compatible with older versions.
``` emacs-lisp
(when (version< org-version "9.2")
  (defalias 'org-set-tags-to 'org-set-tags))
```

PR #130, opened 2019-01-20
:group 'org-journal is not required in the file defgroup org-journal is defined

cslux:
Refactoring

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -103,8 +103,7 @@ org-journal. Use org-journal-file-format instead.")
 
 (defface org-journal-highlight
   '((t (:foreground "#ff1493")))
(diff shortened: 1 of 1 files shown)


bastibe:
Cool, thank you! Could you point me to some documentation for this?

cslux:
Sure!

(info "(elisp)Variable Definitions")

```
     If a ‘defcustom’ does not specify any ‘:group’, the last group
     defined with ‘defgroup’ in the same file will be used.  This way,
     most ‘defcustom’ do not need an explicit ‘:group’.

```

bastibe:
Wow, cool, thank you very much!

PR #132, opened 2019-01-26
Add journal file encryption support

cslux:
This change will add journal file encryption. I honestly don't see the point of the :crypt: tag if we can just encrypt the hole file. Anyway, I have not removed this functionality. I created a own commit for this as it has nothing to do with the daily/weekly/monthly/yearly journal file functionality I'm currently working on.

best,
Christian

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -93,7 +93,7 @@ org-journal. Use org-journal-file-format instead.")
      "%m" "\\\\(?2:[0-9][0-9]\\\\)"
      (replace-regexp-in-string
       "%Y" "\\\\(?1:[0-9]\\\\{4\\\\}\\\\)" format)))
-   "\\'"))
+   "\\(\\.gpg\\)?\\'"))
 
 ; Customizable variables
 (defgroup org-journal nil
@@ -196,6 +196,11 @@ saves/opens these journal entries, emacs asks a user passphrase
 to encrypt/decrypt it."
   :type 'boolean)
 
+(defcustom org-journal-encrypt-journal nil
+  "If non-nil, encrypt journal files using gpg.
+The journal files will have the file extension .gpg"
+  :type 'boolean)
+
 (defcustom org-journal-encrypt-on 'before-save-hook
   "Hook on which to encrypt entries. It can be set to other hooks
   like kill-buffer-hook. "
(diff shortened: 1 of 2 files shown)


org-journal.el
```diff
@@ -273,7 +278,10 @@ Otherwise, date ascending."
 If no TIME is given, uses the current time."
   (file-truename
    (expand-file-name
-    (format-time-string org-journal-file-format time)
+    (format-time-string
+     (concat org-journal-file-format
```
bastibe: You could write the `(or (and org-journal-encrypt-journals ".gpg") "")` more succinctly as `(if org-journal-encrypt-journals ".gpg" "")`. (Or am I missing something?)

org-journal.el
```diff
@@ -582,7 +590,8 @@ If the date is in the future, create a schedule entry."
   "Read an entry for the TIME and either select the new
   window (NOSELECT is nil) or avoid switching (NOSELECT is
   non-nil."
-  (let ((org-journal-file (org-journal-get-entry-path time)))
+  (let ((org-journal-file (concat (org-journal-get-entry-path time)
+                                  (or (and org-journal-encrypt-journals ".gpg") ""))))
```
bastibe: see above re: `(or (and`

org-journal.el
```diff
@@ -196,6 +196,11 @@ saves/opens these journal entries, emacs asks a user passphrase
 to encrypt/decrypt it."
   :type 'boolean)
 
+(defcustom org-journal-encrypt-journals nil
```
bastibe: I feel this should be `org-journal-encrypt-journal` (no plural), since org-journal is keeping one journal (but many entries).

org-journal.el
cslux: Whoops, right! :-D

cslux:
Files ending with .gpg will be treated as encrypted files by the epa library and will be inserted decrypted, hence the search functionality works.

Adopted your requested changes.

cslux:
Arggh, force pushing is actually a bad thing, github cant resovle the requrested changes now. Sorry!

cslux:
I found a problem! When there are gpg encrypted journal files, and you set `org-journal-encrypt-journal` to `nil` the `org-journal-carryover` tries to open the previous file, but fails to do so, as it does not search any longer for `.gpg` journal files. 

I think it always should search for `.gpg` encrypted journal files, shouldn't it? But ignore them if `org-journal-encrypt-journal` is set to `nil`. 

This is there message I get:
``` 
Non-existent agenda file ~/emacs/org/journal/20190126.  [R]emove from list or [A]bort?
```

cslux:
Or try to open it and if it fails print a message and continue.

cslux:
@bastibe Can be merged!

Or is there anything left?

bastibe:
Cool, thank you very much!

PR #136, opened 2019-02-19
Fix #133; Org needs to be loaded

### LINKED ISSUE #133: Error in org-journal due to org-version variable
In emacs 26.1 after installing org-journal from melpa using package installer emacs throws an error:

```
Symbol's value as variable is void: org-version
```

In the code in org-journal.el file it has a code which results in error:

```
62  ;;; Code:
63  (when (version< org-version "9.2")
64    (defalias 'org-set-tags-to 'org-set-tags))
```
I think error is with org-version and org-journal for emacs 26.1.

cslux:


bastibe:
Thank you!

PR #139, opened 2019-02-26
Fix #138

### LINKED ISSUE #138: When save journal entry it ask gpg key and related stuff
Latest version from melpa at 25 feb ask me choice or enter gpg key when perform save - I don't need encryption and don't use it in org-journal. 

Please disable all this stuff by default. Thanks.

Forced to install previous version - elpa/org-journal-20190121.909. This version works as expected.

cslux:
This fixes a bug introduced in commit b0f3d326323c491aebcdd0465952e49d33f0346d.

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -280,9 +280,9 @@ Otherwise, date ascending."
 If no TIME is given, uses the current time."
   (let ((file (file-truename
                (expand-file-name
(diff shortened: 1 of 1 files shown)


bastibe:
Thank you for your quick response! I had looked at your code, but didn't notice the problem, either.

PR #145, opened 2019-03-15
Some improvements and refactoring

cslux:
@bastibe Can review my changes please?

Thanks!

Christian

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -133,7 +133,7 @@ org-journal. Use `org-journal-file-format' instead.")
 (defcustom org-journal-dir "~/Documents/journal/"
   "Directory containing journal entries.
 
-Setting this will update theinternal `org-journal-file-pattern' to a regex
+Setting this will update the internal `org-journal-file-pattern' to a regex
 that matches the directory, using `org-journal-dir-and-format->regex', and
 update `auto-mode-alist' using `org-journal-update-auto-mode-alist'."
   :type 'string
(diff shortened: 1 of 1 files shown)


bastibe:
As far as the code goes, this looks good! 

If I remember correctly, the reason for the `(calendar-basic-setup nil t)` and `(calendar-exit)` was that `calendar-date-compare` was not available unless the calendar was opened at least once. Do your changes work on a newly started Emacs without having displayed the calendar?

Does this mean that calendar entries are no longer marked by default? If this is a performance problem for you, I would suggest making it an option. It doesn't seem to be an issue on my machine.

cslux:
> If I remember correctly, the reason for the `(calendar-basic-setup nil t)` and `(calendar-exit)` was that `calendar-date-compare` was not available unless the calendar was opened at least once. Do your changes work on a newly started Emacs without having displayed the calendar?

Works! Maybe it was due to the missing `(require 'calendar)` which gets called now by `(require 'cal-iso)`. I tested with `emacs -Q`.

> Does this mean that calendar entries are no longer marked by default? If this is a performance problem for you, I would suggest making it an option. It doesn't seem to be an issue on my machine.

As soon as you open the calendar the entries get marked.

bastibe:
OK, cool! Thank you! 
Code maintenance is one of the most thankless, yet necessary parts in a project's life cycle.

PR #147, opened 2019-03-20
Adds Badges for Github issues and such

bastibe:
Adds a few badges to the readme. I like these for my own use to get an overview of current activity within my own projects.


cslux:
@bastibe Looks good!

PR #148, opened 2019-03-22
Add support for tag settings specific to org-journal

### LINKED ISSUE #109: Set default tags for journal entries? (like with org-tag-alist)
You know how with org-tag-alist, you can [establish a default set of tags](http://orgmode.org/manual/Setting-tags.html#Setting-tags), and then insert them easily with org-set-tags-command? I’m wondering how to do that in a way that’s specific to org-journal. (I’m guessing you could find some hack-y way to add a “#+TAGS:” buffer-local variable before or after the datestamp of every entry, but I’d like to look for a more elegant solution that doesn’t clutter up my entries, first.) Thanks!

TinaRussell:
Fixes #109

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -243,6 +243,23 @@ See agenda tags view match description for the format of this."
   "When :desc, make search results ordered by date descending, otherwise date ascending."
   :type 'symbol)
 
+(defcustom org-journal-tag-alist nil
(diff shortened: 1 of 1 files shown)


cslux:
@TinaRussell If you like, you can resolve it yourself, but I'm also willing to fix it.

PR #159, opened 2019-05-16
[doc] Fix Spacemacs org-journal support 404 link

duianto:
And change the emphasis from ` (grave accents) to = (equal signs).

---

The issue was caused by the url being double encoded as explained in this SO question:
https://stackoverflow.com/questions/13968282/unknown-characters-252b-in-url

>The URL has been double encoded. %25 is the escape sequence for "%", so a regular %2B got escaped again to %252B.

source: https://stackoverflow.com/a/13968361

### DIFF
--- a/README.org
+++ b/README.org
@@ -357,7 +357,7 @@ variables.
 
 Yes you can!
 
-- To use =org-journal= with Spacemacs from master branch, you must do this:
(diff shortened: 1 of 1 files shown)


bastibe:
Cool, thank you!

PR #161, opened 2019-05-31
NEW Defines custom `search-forward`.

vitorqb:
Adds a custom variable `org-journal-search-forward-fn`, which is
used in `org-journal-search-do-search` to search forward in the
buffer for a match. This is usefull if you want (like I do) to search
the journal using regexp instead of string matching.


cslux:
Thank you @vitorqb, merged!

PR #174, opened 2019-10-11
Create the directory for org-journal-cache-file

akirak:
This patch ensures that the parent directory of `org-journal-cache-file` is created before the file is saved.

It also throws an error if the file is still unwritable.

This was originally pointed by @tarsius while I was trying to add this package to `no-littering`:

https://github.com/emacscollective/no-littering/pull/101


cslux:
Thanks @akirak, merged!

PR #180, opened 2020-01-01
Set the customization type of org-journal-dir to directory

akirak:
`org-journal-dir` currently has `string` type, but it would be slightly better to make it a `directory`, which is defined as follows in the info:

> ‘directory`
>      The value must be a directory.  The widget provides completion.


cslux:
Thank you @akirak!

Merged!

PR #183, opened 2020-01-09
Update README.org

whiskeyputers:
Reword "Some key-bindings in org-journal overwrite org-mode key bindings" to improve clarity.

### DIFF
--- a/README.org
+++ b/README.org
@@ -383,11 +383,20 @@ Yes you can!
 
 *** Some key-bindings in org-journal conflict with org-mode key bindings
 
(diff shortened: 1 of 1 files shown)


bastibe:
Thank you for your pull request! This has clarified one point for me: there are actually *two* conflicts here, not one.

1. some org-journal key bindings don't work from regular org-mode buffers
2. some org-mode key bindings don't work in org-journal buffers

The current README addresses 2, but not 1. Your revised README addresses 1, but not 2. I think we should mention both issues, actually.

Would you like to incorporate this into your pull request?

whiskeyputers:
Just committed [b3a82f4](https://github.com/whiskeyputers/org-journal/commit/b3a82f438673ba9c86d6e7f588e4d1b0c0afb28d) to address both scenarios.

bastibe:
That's really cool, thank you very much!

PR #184, opened 2020-01-10
Add org-journal-file-header

### LINKED ISSUE #182: Adding top-of-file metadata?
I've been creating daily journals for a while now, but in practice, I create a summary for every week. On the one hand, that's a good opportunity to synthesize the work of the week. On the other hand, I wouldn't have to cut and paste quite so much if I just kept a weekly journal.

So I'm thinking of switching to weekly.

But if I did that, I think I'd want to have a preamble:

```
#+TITLE: Some title, probably generated by a format-time-string or a function
#+OPTION: some option
#+HEADER: maybe other headers

* 6 January 2020
** 09:00 stuff I did
** 12:00 more stuff I did
* 7 January 2020
** 07:30 yet more stuff
...
```

On a casual inspection of the code, it didn't seem like this was configurable today. It's also not entirely clear if the extra headers would introduce any problems. I don't think they would, but...

Is this something you think could be made to work?

ndw:
Fix #182 

This PR adds an `org-journal-file-header` string that is inserted at the top of new journal entries. If the value is a function, the result of evaluating the function is inserted. FWIW, here's my experimental configuration for weekly journals:

```
(defun ndw/org-journal-monday (&optional time)
  "Return the TIME cast back to the previous Monday.
If TIME isn’t specified, the (current-time) is assumed.
If today is Monday, return today."
  (let* ((now    (if time time (current-time)))
         (dow    (string-to-number (format-time-string "%u" now)))
         (delta  (if (= dow 1)
                     0
                   (* (- dow 1) (* 24 3600)))))
    (seconds-to-time (- (time-to-seconds now) delta))))

(customize-set-variable 'org-journal-file-header
   (lambda ()                        
     (let ((monday (ndw/org-journal-monday)))
       (insert "#+TITLE: ")
       (insert (format-time-string org-journal-date-format monday))
       (insert "\n#+STARTUP: showeverything\n#+FILETAGS: :Journal:\n\n"))))
```

### DIFF
--- a/README.org
+++ b/README.org
@@ -217,6 +217,8 @@ Customization options related to the journal file contents:
 
 - =org-journal-time-prefix= - a string that will prefix every entry within a daily journal file.
 
+- org-journal-file-header - a string that will be inserted at the top of every new journal file.
+
 *** An example setup
 
 A very basic example of customization.
--- a/org-journal.el
+++ b/org-journal.el
@@ -311,6 +311,11 @@ search works with regexps."
   "If `t', journal entry dates will be cashed for faster calendar operations."
   :type 'boolean)
 
+(defcustom org-journal-file-header nil
+  "If non-nil, contents will be inserted at the top of new journal files.
+If you define it as a function, it is evaluated and inserted."
+  :type 'string)
+
 (defvar org-journal-after-entry-create-hook nil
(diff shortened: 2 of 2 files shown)


cslux:
Integrated into master.

PR #186, opened 2020-01-24
Allow changing of weekly journal start day.

Spauldo:
Added two customization options org-journal-start-on-weekday
and org-journal-use-agenda-start-on-weekday.  The first changes
the start day of daily journals.  The second tells org-journal
to use the value of org-agenda-start-on-weekday for its start
day.  When the second option is set, the first is ignored.

Do note that established journals will no longer work when the
start day is changed.

Please excuse the indentation bits that were sucked into the diff
(lines 82-89).  Apparently aggressive-indent-mode took issue with
some of the indentation I was scrolling through.  I can create a new
pull request without those, if you like.

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -79,14 +79,14 @@
 From branch \"emacs-26\", added for compatibility.
 "
     (cond
-      ((null alist2) alist1)
-      ((null alist1) alist2)
-      (t (let ((alist2-cars (mapcar (lambda (x) (car-safe x)) alist2))
-               to-add)
-           (dolist (i alist1)
-             (unless (member (car-safe i) alist2-cars)
-               (push i to-add)))
-           (append to-add alist2)))))
+     ((null alist2) alist1)
+     ((null alist1) alist2)
+     (t (let ((alist2-cars (mapcar (lambda (x) (car-safe x)) alist2))
+              to-add)
+          (dolist (i alist1)
+            (unless (member (car-safe i) alist2-cars)
+              (push i to-add)))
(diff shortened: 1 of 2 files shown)


org-journal.el
```diff
@@ -158,6 +158,31 @@ to be done manually by calling `org-journal-invalidate-cache'."
           (const :tag "Monthly" monthly)
           (const :tag "Yearly" yearly)))
 
+(defcustom org-journal-start-on-weekday 1
+  "What day of the week to start a weekly journal.
+
+When org-journal-file-type is set to 'weekly, start the week on
+this day.  Defaults to Monday.
+
+Not that when org-journal-use-agenda-start-on-weekday is non-NIL,
```
cslux: The last sentence can than obviously be removed :)

org-journal.el
```diff
@@ -79,14 +79,14 @@
...
     (cond
-      ((null alist2) alist1)
-      ((null alist1) alist2)
-      (t (let ((alist2-cars (mapcar (lambda (x) (car-safe x)) alist2))
-               to-add)
-           (dolist (i alist1)
-             (unless (member (car-safe i) alist2-cars)
-               (push i to-add)))
-           (append to-add alist2)))))
+     ((null alist2) alist1)
```
cslux: I will fix all the indentation before merging it.

Spauldo:
Sounds fine to me.  Give me a bit and I'll make the changes.

Spauldo:
Should be good now.

PR #212, opened 2020-03-29
Use new style created timestamp (Resolve #188)

### LINKED ISSUE #188: Use inactive timestamp for created property
or perhaps the date part of `org-journal-date-format`. It's hardcoded to be %Y%m%d. My personal preference would be the ISO format or %F but perhaps it should be a customisable option. There also seems to be two spaces between the property keyword and its value. Is that necessary?

cslux:
From new on inactive timestamp will be used for the created property.
The old style is still supported and will not be converted.

### DIFF
--- a/org-journal-test.el
+++ b/org-journal-test.el
@@ -51,31 +51,29 @@
 
 (ert-deftest org-journal-calendar-date-from-file ()
   "Should return a list with day/month/year"
-  (org-journal-dir-test-setup)
-  (let* ((org-journal-dir org-journal-dir-test)
-         (org-journal-file-pattern (org-journal-file-pattern-test)))
-    (should (equal (org-journal-file-name->calendar-date
-                    (expand-file-name "20190103" org-journal-dir))
-                   '(1 03 2019)))))
+  (org-journal-test-macro
+   (should (equal (org-journal-file-name->calendar-date
+                   (expand-file-name "20190103" org-journal-dir))
+                  '(1 03 2019)))))
 
 (ert-deftest org-journal-convert-time-to-file-type-time-test ()
   "Testing"
-  (let ((time (current-time))
-        (org-journal-file-type 'daily))
-    (should (equal (org-journal-convert-time-to-file-type-time time)
-                   time))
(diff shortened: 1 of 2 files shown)


cslux:
This leaves the existing created property untouched, or should I convert it to the new style as soon as the old style gets detected. This would have the advantage that we could remove the old code somewhere in the future (after one year or so?).

@doolio you could try this if you like

bastibe:
This looks good to me!

Three specific comments:
- Perhaps you could add a comment to the relevant sections, explaining why we are supporting two different formats.
- It might also be worthwhile to make the timestamp format a configurable variable, mostly because the new style contains localized words that might trip up some use cases.
- It might be good to include both kinds of timestamps in the test, just to make sure we don't accidentally break people's existing data in the future.

But, these are details. Overall, this seems a worthwhile change.

cslux:
@bastibe I have improved it, and fixed some other small bug etc.

Please review again, thanks.

Christian

cslux:
> By the way, do you know if there is a way to set up continuous integration with those?

Sure, I did it for rtags too (https://github.com/Andersbakken/rtags), where I'm also a maintainer. I'll add CI using travis or something the like. I need to do a little research how other emacs-lisp projects are doing it.

> As for the rest of the changes, my only comment is on the two choices we get, `'ymd` and `'inactive`. I see why there are these two specific choices. However, they do beg the question why we are restricting them to these particular choices, as opposed to an arbitrary format string, and a translation layer similar to `org-journal-dir-and-format->regex`.

I'm with you, it doesn't feel clean and sane. I thought about the same implementation. But than we need to verify that it includes %Y, %m and %d, at least. I'll try to come up with another solution were only a defcustom of type string is required.

cslux:
The thing is, for weekly, monthly and yearly journal files, we rely on an unique timestamp. From this timestamp, we create a date (Month Day Year), from that date we need to be able to reconstruct the created timestamp to find it again. So in theory, the user can use any time format placeholder which satisfy this requirement.

cslux:
Merged into master, closing!

PR #217, opened 2020-04-05
Add org-journal-after-header-create-hook

nivekuil:
Add a hook that runs after the top-level date entry is created. I wanted a hook here to better add properties to the top-level drawer, e.g. a TIMEZONE property like
```
(add-hook 'org-journal-after-header-create-hook
                 (lambda ()
                   (org-set-property "TIMEZONE" (format-time-string "%Z"))))
```
This could be done with the existing after-entry-create-hook, but aside from being less efficient, the entry hook doesn't run if a journal entry is added through org-capture, like in the example given in the README, without further abuse.  Additionally there are likely other things that are best done only once per date.

### DIFF
--- a/README.org
+++ b/README.org
@@ -335,7 +335,7 @@ With this done, you can export your agenda, including your scheduled entries, wi
 *** Journal Capture Template
 
 You can configure a capture template in order to integrate =org-journal= with =org-capture=,
-as in the following example:
+as in the following example for a daily journal:
 
 #+BEGIN_EXAMPLE emacs-lisp
   (defun org-journal-find-location ()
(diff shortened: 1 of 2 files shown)


cslux:
Hey @nivekuil! Thank you for your PR. Looks good to me. @bastibe Please have look at this PR.

bastibe:
Nice work, @nivekuil! Would you like to add a short section to the README, maybe with an example usage?

nivekuil:
Sure, added a section describing the hooks. Unrelatedly, I also clarified the org-capture example to note that it only works for daily journal files.

bastibe:
Cool! I like it!

@cslux, are you OK with merging this?

cslux:
Yes!

bastibe:
Well than, @nivekuil, thank you very much for your contribution!

PR #232, opened 2020-05-16
Add org-journal-open-current-journal-file interactive function

psibi:
This allows me to open current journal file easily.

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -1036,6 +1036,14 @@ If no next/previous entry was found print MSG."
       (message msg)
       nil)))
 
+(defun org-journal-open-current-journal-file ()
+  "Open the current journal file"
+  (interactive)
+  (let ((org-journal-file (org-journal-get-entry-path)))
+    (if (file-exists-p org-journal-file)
+        (funcall org-journal-find-file org-journal-file)
+      (message "Journal file %s not found" org-journal-file))))
+
 (defun org-journal-open-next-entry (&optional no-select)
   "Open the next journal entry starting from a currently displayed one."
   (interactive)
--- a/org-journal.el
+++ b/org-journal.el
@@ -981,6 +981,15 @@ arguments (C-u C-u) are given. In that case insert just the heading."
       (save-excursion
         (insert "\n<" scheduled-time ">")))))
 
+(defun org-journal-goto-entry (date)
+  "Goto DATE entry in current journal file."
+  (widen)
+  (goto-char (point-min))
+  (if (org-journal-daily-p)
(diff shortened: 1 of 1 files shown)


psibi:
@bastibe When I have multiple buffers open, I use this function frequently to go the current journal file. Is there any existing function which can be used to do that (I actually went briefly through the existing functions, before just writing this out) ?

bastibe:
I usually use `org-journal-new-entry` with a prefix in this situation, i.e. `C-u C-c C-j`.

psibi:
@bastibe Using that function will interfere with the variable `org-journal-skip-carryover-drawers` and will deletes the drawers if it is set (which in my case is set) and hence cannot be used in my current workflow.

cslux:
> @bastibe Using that function will interfere with the variable `org-journal-skip-carryover-drawers` and will deletes the drawers if it is set (which in my case is set) and hence cannot be used in my current workflow.

That sounds like a bug!

@bastibe I think @psibi wants to just open the current journal file without creating a new heading/entry. org-journal-new-entry called with a prefix argument doesn't insert an entry, but does insert a heading though. 

We could also supercharge org-journal-new-entry with two prefix arguments. If two prefix args are given just open the current journal file, and if it doesn't exists, create it. What do you think @psibi @bastibe?

psibi:
> @bastibe I think @psibi wants to just open the current journal file without creating a new heading/entry. org

Yeah, that's right.

> We could also supercharge org-journal-new-entry with two prefix arguments. If two prefix args are given just open the current journal file, and if it doesn't exists, create it. What do you think @psibi @bastibe?

Sounds good to me.

bastibe:
> We could also supercharge org-journal-new-entry with two prefix arguments. If two prefix args are given just open the current journal file, and if it doesn't exists, create it. What do you think @psibi @bastibe?

Honestly, most of the time I actually just want to open the most recent journal entry without adding anything new just yet. For this application, sticking with @psibi's PR is actually more convenient than adding yet another modifier to `org-journal-new-entry`. (Although we might want to expand the PR to open the *most recent* entry instead of *today's*).

cslux:
> Honestly, most of the time I actually just want to open the most recent journal entry without adding anything new just yet. For this application, sticking with @psibi's PR is actually more convenient than adding yet another modifier to `org-journal-new-entry`. (Although we might want to expand the PR to open the _most recent_ entry instead of _today's_).

Sounds good to me.

cslux:
Do we want to bind it to a key, or should we left it to the user?

bastibe:
Looks good to me!

PR #240, opened 2020-05-23
Use magic-mode-alist instead auto-mode-alist to determine the mode

cslux:
In `magic-mode-alist` we can use a function to determine if the
file is a journal file, and return `t` if it is one, otherwise `nil`.
That way we don't need any `:custom` keyword in the defcustom's, and hence,
org-journal is less complex to setup and use.

Takes away the complexity from the user, and moves it to the developer.

@bastibe Please have look at this.


cslux:
- [ ] Update documentation @cslux

bastibe:
I had no idea `magic-mode-alist` existed, but it seems a much better fit for our purposes! Great find, @cslux!

PR #257, opened 2020-06-17
Update use-package example to use :custom

jakemcc:
The function that calculates the directory regex needs to run with `org-journal-dir` is customized. When I had my configuration under `:config` with `setq` this didn't happen. Moving to using `use-package`'s `:custom` feature works.

### DIFF
--- a/README.org
+++ b/README.org
@@ -236,9 +236,9 @@ For users of =use-package=, this setup could look like the following:
 (use-package org-journal
   :ensure t
   :defer t
-  :config
-  (setq org-journal-dir "~/org/journal/"
-        org-journal-date-format "%A, %d %B %Y"))
+  :custom
+  (org-journal-dir "~/org/journal/")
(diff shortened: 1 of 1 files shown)


cslux:
Hey @jakemcc! Did you install org-journal from MELPA or MELPA-Stable?

jakemcc:
Hello @cslux! Looks like I have org-journal-2.0.0, so MELPA-stable. Should I switch to MELPA?

cslux:
Depends on your liking. MELPA-Stable is fine, you would need to select the
tag 2.0.0 to get the correct readme version
(https://github.com/bastibe/org-journal/tree/2.0.0).
We release version 2.1.0 in July. With that version, no :custom is required
anymore.

jakemcc:
@cslux Ahh, I should have thought of looking at the readme for that version. Thanks for letting me know and thanks for the work on org-journal.

Issue #259, opened 2020-06-18
Incomplete org-journal-format->regex conversion with a org-journal-file-format containing "%B"


jinheebaek:
`org-journal-file-format` with "%B" date string no longer works after an update.

I have temporarily modified `org-journal--format-rx-alist` as shown below for my journal files. 

Is it intended not to use %B string? 

(defvar org-journal--format-rx-alist
  '(("%[aA**B**]" . "\\\\(?4:[a-zA-Z]\\\\{3,\\\\}\\\\)")
    ("%d" . "\\\\(?3:[0-9][0-9]\\\\)")
    ("%m" . "\\\\(?2:[0-9][0-9]\\\\)")
    ("%Y" . "\\\\(?1:[0-9]\\\\{4\\\\}\\\\)")))

Thanks!

cslux:
We can add %B do it! Can you create a PR please? There also two function doc-strings which need to be updated when adding %B support -- `org-journal-file-format` and `org-journal-created-property-timestamp-format`.

PR #262, opened 2020-06-22
Support %b,%B representing the month name in org-journal-file-format

jinheebaek:
It adds supports "%b" and "%B" format string in org-journal-file-format representing the name of the month (issue #259)

Thanks


cslux:
Thank you! Merged!

PR #278, opened 2020-07-21
Remove harcoded key-bindings and move them to user-config

### LINKED ISSUE #255: Use C-c j as prefix for org-journal-mode
Some bindings overwrite org-mode bindings like C-c C-s (org-journal-search -> org-schedule).

zaeph:
Closes #255.

I'm not a big fan of hardcoding key-bindings for minor-modes that are magically loaded, especially when they override existing ones.  It makes it complicated for users to rebind or unbind them without first emptying the keymap.

I'd like to move those key-bindings from codebase to user-config.  This is something that we've done with [Org-roam](https://github.com/org-roam/org-roam#installation), and we haven't encountered any problem with it.

I took the liberty to update the `CHANGELOG` as well.

Thank you for `org-journal`. :)

### DIFF
--- a/CHANGELOG
+++ b/CHANGELOG
@@ -127,3 +127,6 @@
   - Lot of general code improvements and bug fixes
   - Documentation update (Thank you, Doolio)
   - Support %b/%B in org-journal-file-format (Thank you, Jinhee Baek)
+
+## TBD
+  - Move default key-bindings from codebase to user-config
--- a/README.org
+++ b/README.org
@@ -422,7 +422,7 @@ When working in an org-journal buffer the following org-mode key bindings are ov
 - =C-c C-b= (=org-backward-heading-same-level=) with =org-journal-open-previous-entry=
 - =C-c C-j= (=org-goto=) with =org-journal-new-entry=
 
-To workaround this, you can use user bindings of the form =C-c ?=, where =?= can be any letter, to call the org-journal functions. This allows you to have a set of keybindings that work the same in org-mode and org-journal buffers. However, this is Emacs, and if you don't like a key binding, change it!
+To workaround this, you can use user bindings of the form =C-c ?=, where =?= can be any letter, to call the org-journal functions. This allows you to have a set of keybindings that work the same in org-mode and org-journal buffers. However, this is Emacs, and if you don't like a key binding, change it in your configuration!
 
 *** Opening journal entries from the calendar are not editable
 
--- a/README.org
+++ b/README.org
@@ -230,6 +230,10 @@ A very basic example of customization.
 (setq org-journal-dir "~/org/journal/")
 (setq org-journal-date-format "%A, %d %B %Y")
 (require 'org-journal)
+(define-key org-journal-mode-map (kbd "C-c C-f") 'org-journal-next-entry)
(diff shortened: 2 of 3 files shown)


zaeph:
I forgot to mention that this change will require existing users to update their config to maintain the default key-bindings.  I don't think there's any good way to make the change seamless for them, and creating a warn-on-update setup would definitely be overkill.

bastibe:
I am incredibly torn about this. On the one hand, I agree with you and would like to remove the key bindings. Making org-journal a major mode was a bad idea to begin with, and setting fixed key bindings was equally mis-guided on my part.

However, org-journal has an astonishingly large number of users, many of whom will be caught off-guard by such a change. And in all my years of working on Open Source Software, I have learned that nothing erodes trust like breaking changes. And there is always fallout that we maintainers have to deal with. Angry emails, issues, messages, maintenance overhead, and more. 

There have been multiple issues that discussed solutions to this problem already. For example, we might make org-journal into an optional minor mode. In fact, it is entirely unnecessary for org-journal to be a mode at all, it could just be a number of functions and hooks.

At any rate, this problem needs a thoroughly documented, strongly communicated solution that goes deeper than the key bindings themselves. It will in all likelihood require a major version change.

Thus, regretfully, I have to reject this pull request for now. But I invite you to discuss and implement a possible restructuring of org-journal with @cslux and me here or in a new issue.

zaeph:
I understand.  I can't think of a better solution for this key-binding problem, so it looks I'm going to have to port your functions into a separate library, then.  On the one hand, it would have been nice to maintain `org-journal` within the ecosystem we're trying to build for Org-roam, but at the same time, I realise that we've been abusing some of the behaviours to make it do our bidding.

I'll make sure to credit you in the future lib.

Thanks for your time, and good luck for the future!

cslux:
@zaeph I'll add a customizable variable to configure the prefix argument. When it is set to an empty string no bindings will than be configured. By default nothing will change for the user.

zaeph:
> @zaeph I'll add a customizable variable to configure the prefix argument. When it is set to an empty string no bindings will than be configured. By default nothing will change for the user.

Splendid!  Something à la `projectile-command-map` would do nicely.  I'm still iffy about setting it to `C-c j` by default, but at least it gives us the possibility to customize it.

cslux:
> Splendid! Something à la [`projectile-command-map`](https://github.com/bbatsov/projectile/blob/bbcf781d3fddb8e00d0dc10eb68bc2528fb409b3/projectile.el#L4674) would do nicely. I'm still iffy about setting it to `C-c j` by default, but at least it gives us the possibility to customize it.

@zaeph I agree, but some users are verylikly used to the current "C-c C-*" key bindings. I didn't change it with @bcb5bb4. You can now set `org-journal-prefix-key` to an empty string, which means, no default bindings. This variable needs to be set before org-journal gets loaded.

zaeph:
Thanks for taking care of this.

Re:
> I understand. I can't think of a better solution for this key-binding problem, so it looks I'm going to have to port your functions into a separate library, then.

org-roam/org-roam/pull/978 is almost finished.  Functionally wise, I've taken a lot of inspiration from `org-journal`, which I'll mention in the header of the lib.  I think you've had a lot of good ideas UX wise, and I wish I'd found out about you before writing my own implementation for journalling.

Best of luck for the future.

Issue #264, opened 2020-06-22
[BUG]: Carryover not working due to date sorting


eccentric-j:
**Describe the bug**
When creating a new journal entry, sometimes tasks from the last entry are not carried over. This appears to be caused by the sorting of the date entries.

**To Reproduce**
1. Create an entry in the past in my case 20200618.org
2. Add a TODO heading with a timestamp under it
2. Create a new entry in the future. In my case 20200622.org

**Expected behavior**
The TODO entry in 20200618.org is carried over to 20200622.org

**Desktop**
 - OS: OS X 10.15.5 (Catalina)
 - Emacs Version 27

**Your Emacs Configuration**
```lisp
(after! org
  (setq
    org-directory                         "~/Dropbox/org"
    diary-file                            (concat org-directory "/diary")
    org-agenda-include-diary              nil
    org-agenda-file-regexp                "\\`[^.].*\\.org'\\|[0-9]+\\.org$"
    org-agenda-timegrid-use-ampm          t
    org-journal-dir                       (concat org-directory "/journal")
    org-journal-enable-agenda-integration nil
    org-journal-file-format               "%Y%m%d.org"
    org-journal-time-format               "%-l:%M%#p"
    org-journal-carryover-items           "TODO=\"TODO\"|TODO=\"STRT\"|TODO=\"HOLD\"")
  (appendq! org-agenda-files (list org-journal-dir)))
```

**Research**
https://github.com/bastibe/org-journal/blob/master/org-journal.el#L1001-L1009
```lisp
(setq dates (cl-loop
                  for date in dates
                  while (calendar-date-compare (list date) (list calendar-date))
                  collect date into result and count t into cnt
                  finally return (if result
                                        ;; Front
                                        `(,@result ,calendar-date)
                                      ;; Somewhere enbetween or end of dates
                                      `(,calendar-date ,@result ,@(nthcdr cnt dates)))))
```

It seems to be sorting the dates incorrectly. I added advice to that function and logged what the collection algorithm above would return:

**Output**
```
dates 
((6 16 2020)
 (6 18 2020)
 (6 11 2020)
 (6 12 2020)
 (6 13 2020)
 (6 14 2020)
 (6 15 2020)
 (6 16 2020))
```

**Solution Strategies**
- Create a separate function for selecting\listing the dates, even if it uses the same sorting algorithm, it at least gives us options
[...]

PR #291, opened 2020-09-05
Create sort-dates func for org-journal--open-entry

eccentric-j:
As per bastibe/org-journal#264 I'm glad the core issue was fixed, however, I still think it would be useful to have a customizable sort function if more complex filtering would be desired.

This pull request creates org-journal-sort-dates to extract the sorting logic from org-journal--open-entry. 

Please let me know what else needs to be done.


bastibe:
This sounds reasonable to me. Let's see what @cslux thinks of it.

cslux:
@eccentric-j Thanks! It's always good to simplify functions by refactoring part of it into an own function. Merged!

PR #293, opened 2020-09-09
Add capture integration example for monthly journal

wehlutyk:
When using capture with a non-daily journal and the example in the readme, new entries are positioned on the date at the top of the file instead of the current date. I think some of the conversation in #50 was about that.

The new snippet in this PR lets you use capture with a monthly journal (I haven't tested yearly or weekly, but they should work too).

Thanks for the great package, and feel free to do what you want with this!

### DIFF
--- a/README.org
+++ b/README.org
@@ -364,6 +364,22 @@ as in the following example for a daily journal:
                                  "* %(format-time-string org-journal-time-format)%^{Title}\n%i%?")))
 #+END_EXAMPLE
 
+For a monthly journal, replace the definition of `org-journal-find-location` with the following to make sure your entries find the right date:
+
+#+BEGIN_EXAMPLE emacs-lisp
+  (defun sl/org-journal-find-location ()
+    ;; Open today's journal, but specify a non-nil prefix argument in order to
+    ;; inhibit inserting the heading; org-capture will insert the heading.
+    (org-journal-new-entry t)
+    ;; Position point on the journal's today date so that org-capture
+    ;; will add the new entry as a child entry.
+    (goto-char (point-min))
+    (let ((date (format-time-string org-journal-date-format)))
+      (if (not (re-search-forward date nil t))
+          (error "No match for target date in journal file")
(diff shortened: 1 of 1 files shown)


cslux:
Thanks for the PR. I haven't tried your snippet, and I don't think it is really necessary. If I understand it correctly, you would like to insert `org-capture` templates into org-journal files. There are at least 2 possibilities to achieve that very easily.

1. Call `org-capture` with the prefix argument 0
   This would insert the template at point. `C-u M-x org-journal-new-entry RET C-0 M-x org-capture RET`
   You can create a function to combine this, and create a key binding.
    ``` emacs-lisp
    (defun my-org-journal-insert-org-capture ()
       (interactive)
       (org-journal-new-entry t)
       (org-capture 0))
    (eval-after-load "org-journal" (define-key org-journal-mode-map (kbd "C-c j t") 'my-org-journal-insert-org-capture))
    ```
2. Using `org-refile`
    Add the org-journal directory/files to `org-agenda-files` so `org-refile` finds them.
    If your files don't end with ".org" you either need to add all org-journal files to `org-agenda-files` 
    or you have to modify `org-agenda-file-regexp`.
    You would still need to create the org-journal entry first.

wehlutyk:
My description wasn't very clear. I want to use `org-capture` to create a new journal entry, like described in [the existing readme](https://github.com/bastibe/org-journal#journal-capture-template).

But the existing snippet in the readme works only if `org-journal` is configured for daily files. With monthly files, new journal entries created using the existing `org-capture` template always get positioned on the first day of the month, instead of on the current day (because `(goto-char (point-min))`). So the snippet in this PR lets `org-capture` find the 'today' heading in the journal.

wehlutyk:
I should have started an issue for this beforehand :smile:

cslux:
Make sense. Merged but modified it a bit. I think it is sufficient to narrow to the subtree.

cslux:
Thanks!

PR #307, opened 2020-10-28
Enable to costomize the way to handle the old carryover entries

june3474:
**Feature Request**
  When carryover occurs, old carryover entries in the previous day's journal are deleted. No other options.
  It would be nice if users can choose what they do with old carryover entries.

**Solution I would like**
   To make it enable for users to choose a function to handle the old entries instead of removing them all.
   The function could default to deleting old entries.


june3474:
There was a mistake with the example in README.org.
I will make a new pull request after fixing it.

cslux:
Thanks for this PR. Looks good! Merged!

PR #314, opened 2020-12-17
Solve error when org-journal-dir has symlink in the path

emiltoacs:
Solve the error "Failed to extract year from file: %s" for journal file with
link in their path.

Inside function org-journal--file-name->calendar-date there is a mismatch
between ‘file-pattern’ and the parameter ‘file-name’. Because file-pattern has
its path with symlink resolved with file-truename but the parameter file-name
still have the path with symlink.

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -964,23 +964,24 @@ Month and Day capture group default to 1."
   (let ((file-pattern (org-journal--dir-and-file-format->pattern))
         (day 1)
         (month 1)
-        year)
+        year
+        (file (file-truename file-name)))
     (setq year (string-to-number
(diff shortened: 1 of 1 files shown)


bastibe:
Thank you for your contribution!

Is this truly the only place where the symlink causes a problem? This type of thing usually causes issues at more than one spot.

emiltoacs:
It seems to me that it's the only place where true-filename causes problem when compared with a filename directly but I just did a quick review of the code and as far I can tell, in my everyday usage of org-journal it works well.

bastibe:
Thank you very much, then!

Issue #185, opened 2020-01-23
Make Weekly Journal Start on a Different Day


Spauldo:
Currently, `org-journal` creates weekly journal files that start on Monday.

I am looking to use `org-journal` for my weekly reports.  When I'm on site, I have to turn in a document to the government that gives my activities from Sunday - Saturday.  I am planning on creating an export filter for this.  It'd be a lot easier if I could have weekly journal files that run Sunday - Saturday.

I've dug around in the code and found the function `org-journal-convert-time-to-file-type-time`, which seems to have chosen Monday because of the eccentricities of the ISO calendar.  I figure adding a number between -1 and 5 to the value returned by `calendar-iso-to-absolute` will accomplish what I need to do.  That number could possibly be linked to `org-agenda-start-on-weekday` as well, as I also plan to have my agenda start on Sunday.

I'm planning on doing this for my own needs regardless, but would you be interested in a pull request?  Do you see any hidden gotchas I might run into?

bastibe:
I would absolutely be interested in a pull request! I would guess that there is a variable for setting the week-start somewhere in org-mode, and we should probably reuse that.

Spauldo:
Pull request #186 added.  Let me know if you want changes to it.  It was a bit more complicated than I would have liked due to the way ISO dates work.  I've done a bit of testing and it seems to work, but I didn't have time for extensive testing.

aragaer:
Looks like it doesn't work properly when weekday is sunday.
I have a file named journal-2020-10-18.org (that was sunday) with a Sunday entry and CREATED: 20201018
When I try to create a new entry on monday 19th, I'm getting the following error:
```
org-journal--search-forward-created: Search failed: "[ 	]*:CREATED:[ 	]+20201018[ 	]*$"
```
Deleting the Sunday entry from the file allows me to add monday entries.

Spauldo:
I tested it and confirmed it doesn't seem to work on Sundays.

I stopped using this shortly after I did the pull request because my
reporting requirements changed.  I might be able to look into this at some
point, but at the moment I'm swamped with my job.

I'd suggest reverting the pull request for now.

PR #322, opened 2021-01-17
Fix "start on sunday"

aragaer:
Trying to fix the issue in #185

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -139,13 +139,13 @@ to be done manually by calling `org-journal-invalidate-cache'."
 When `org-journal-file-type' is set to 'weekly, start the week on
 this day.  Default is Monday."
   :type '(choice
-	  (const :tag "Sunday" 0)
-	  (const :tag "Monday" 1)
-	  (const :tag "Tuesday" 2)
-	  (const :tag "Wednesday" 3)
-	  (const :tag "Thursday" 4)
-	  (const :tag "Friday" 5)
-	  (const :tag "Saturday" 6)))
+          (const :tag "Monday" 1)
+          (const :tag "Tuesday" 2)
+          (const :tag "Wednesday" 3)
+          (const :tag "Thursday" 4)
+          (const :tag "Friday" 5)
+          (const :tag "Saturday" 6)
+          (const :tag "Sunday" 7)))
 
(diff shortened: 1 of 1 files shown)


cslux:
Thanks for your contribution. Do have an example journal file, and configuration to share which reproduce this issue?

aragaer:
My configuration is
```emacs-lisp
  :custom
  (org-journal-file-type 'weekly)
  (org-journal-date-format "%A, %Y-%m-%d")
  (org-journal-start-on-weekday 7)
```
Today (2020-01-17, Sunday) I tried to create an entry for today, but I already had a TODO item in journal for tomorrow (2020-01-18, Monday). With the code above I've managed to do that.
```
#+startup: overview
#+startup: shrink
#+category: journal
* Monday, 2021-01-18
:PROPERTIES:
:CREATED:  20210118
:END:
** TODO Something
SCHEDULED: <2021-01-18 Mon>
```

cslux:
I was able to reproduce it with a test, and indeed 0 is wrong and 7 is correct. When 0, the journal entry would be added to the journal file 20210110, instead of the correct one 20210117.

Can you remove the second commit (https://github.com/bastibe/org-journal/pull/322/commits/5ca5b2f24a4e51fcf53552987e53df95ea5ec2d1) from this PR please. I don't see why these changes are necessary. 

Also please move sunday to the end of the list (numerical sorted).

aragaer:
Second commit is not directly related to this, it's about inserting the new entry into existing file at the correct position. Current implementation gathers the list of all already present entries in descending order (starting from future ones and moving to past ones) and tries to find a first entry that is "older" than new. If that succeeds, new entry is inserted after such entry. However if all entries are "younger" than the new inserted one, no such entry matches and new entry is inserted in the end. In example I provided, Sunday entry would be added at the end of file, after the Monday entry (it should be the other way). My second commit fixes exactly that issue.

I can submit a different PR for that if necessary.

cslux:
It should indeed be added before Monday when start of weekday is set to
Sunday. Please create a new PR for that. I'll create a test case for it which
proofs the current implementation is wrong.

cslux:
Appreciate your help! Merged.

PR #327, opened 2021-02-03
Fix incorrect entry-path when extend-today

yangsheng6810:
When we are beyond midnight but within org-extend-today-until, we should
create entry in yesterday's file, not today's. This bug was introduced
in commit 49c55913ac12d5d30ff9667fc51759b418ff7fc8


cslux:
Thanks! There is obviously a test missing for that case. Merging!

Issue #328, opened 2021-02-03
We should use find-buffer-visiting instead of get-file-buffer


wangwb98:
**Describe the bug**

Recently I set directory-abbrev-alist since my "~/org" folder and "~/sync_folder/org" folder are in fact same directory (the first one is only a symbolic link to the other). So I set directory-abbrev-alist like below, thus emacs can always use /home/test/org path.

 (setq directory-abbrev-alist
   '(("\\`/home/test/sync_folder/org" . "/home/test/org")))

Then I found org-journal will close the target buffer strangely, when it's doing org-journal-open-entry. Later I find out that its "get-file-buffer" call didn't return the already opened buffer, due to it's too strict about the file name expansion. See https://www.gnu.org/software/emacs/manual/html_node/elisp/Buffer-File-Name.html for difference between get-file-buffer  and find-buffer-visiting.

I think we'd better use find-buffer-visiting instead since it can tolerate the file name expansion.

**To Reproduce**
enable directory-abbrev-alist for org journal directory.

**Expected behavior**
In my set up, a buffer for ~/sync_folder/org/test.org is exactly the same one for a buffer for ~/org/test.org.
After I replace all the get-file-buffer, it works fine now in my place. Do you want me to generate a PR for this?

**Screenshots**
N/A since ithe buffer is simply killed.

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 - Emacs Version 28

**Your Emacs Configuration**
The part relevant to org-journal.

cslux:
According to documentation this change shouldn't impact us in any way. So yes please. I PR would be appreciated.

PR #329, opened 2021-02-04
Replace get-file-buffer with find-buffer-visiting

wangwb98:
Since find-buffer-visiting works better with file name expansion, based on the
description of those two functions.

It helps for people whose directory-abbrev-alist value include org-journal-dir.


wangwb98:
fixed the issue #328

cslux:
Thanks, merged!

PR #335, opened 2021-02-20
Create insert journal headline at point

dalanicolai:
I have a project where I would like to use journal style headlines outside a journal file. This small adjustment enables the option to insert journal style headlines at point. I thought it might be useful sometimes for other user too. If you think the extra command is too "exclusive" to be added by default then it is no problem if you remove it. But then it would still be nice to keep the optional `insert-at-point` argument.

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -632,7 +632,7 @@ This allows the use of `org-journal-tag-alist' and
       (when (re-search-backward "^#\\+" nil t)
         (org-ctrl-c-ctrl-c)))))
 
-(defun org-journal--insert-entry-header (time)
+(defun org-journal--insert-entry-header (time &optional insert-at-point)
   "Create new journal entry if there isn't one."
   (let ((entry-header
          (if (functionp org-journal-date-format)
@@ -641,7 +641,8 @@ This allows the use of `org-journal-tag-alist' and
              (user-error "org-journal-date-format is empty, this won't work"))
            (concat org-journal-date-prefix
                    (format-time-string org-journal-date-format time)))))
-    (goto-char (point-min))
+    (unless insert-at-point
+      (goto-char (point-min)))
     (unless (if (org-journal--daily-p)
                 (or (search-forward entry-header nil t) (and (goto-char (point-max)) nil))
               (cl-loop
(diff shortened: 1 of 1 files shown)


bastibe:
I like it! Let's see what @cslux will say.

bastibe:
On a second thought, and a more thorough look at your code, I feel that this behavior does not fit well into org-journal. Reading through org-journal's code, it is not clear how this functionality is related to org-journal's behavior.

I think I would advise you to maintain a fork or a derivative project for this functionality.

That said, I would have no problem with adding an optional argument or configuration variable here and there for enabling your fork.

dalanicolai:
Okay thanks, that is fine also. On the other hand, all arguments I've added are optional, and should not break anything.

The only parts that could break something would be
- moving the TODO in front of the time in the headline (in #338), but as I wrote in that PR (with a link to org's documentation), that is required for org to handle the HEADLINE as a todo.
- adding the time within the time-stamp (but that was merged already by cslux, he (or me) just had to update a single test)
- inserting the scheduled keyword when creating a scheduled entry

I would say these are nice little "improvements/corrections" (if indeed they do not break the functionality), but they are not essential at all of course. Anyway, no problem for me to maintain a fork.

bastibe:
I saw that all the arguments are optional. My concern is more with the maintenance surface. Optional arguments only make sense if the code itself is actually using them--otherwise, how would you decide whether to delete some apparently unused piece of code?

Therefore, such affordances for third parties must be kept to a minimum, and rigorously documented (and tested, if possible). In case of the insert-at-point arguments, I felt that they needed to be introduced in a few too many places for comfort. Perhaps there is a way of refactoring org-journal's internals such that insertion and location are handled separately, and can be reused by your code more easily (just thinking out loud here). That would be a preferred solution to me.

As for the order of date and TODO, as well as the scheduled keyword, I am all in favor of those.

All of that said, I am not really the maintainer of org-journal any longer. If @cslux later decides that this pull request is acceptable after all, I am entirely ok with that, too.

ieugen:
Would love to see this functionality merged. 
I have my regular org journal file but I would like to keep a journal for work related stuff in separate files so I have a track record. 

However, org-journal does not make that easy at all.

PR #337, opened 2021-02-24
Add scheduled entry time of the day

dalanicolai:
Probably this `t` was forgotten here. This function asks for time-stamp format, For sure when users include a time of the day there, they would like to have it included in the timestamp.

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -1067,7 +1067,7 @@ With non-nil prefix argument create a regular entry instead of a TODO entry."
       (insert "TODO "))
     (save-excursion
       (insert "\n")
-      (org-insert-time-stamp time))))
+      (org-insert-time-stamp time t))))
 
 ;;;###autoload
 (defun org-journal-reschedule-scheduled-entry (&optional time)


cslux:
> Probably this `t` was forgotten here. This function asks for time-stamp format, For sure when users include a time of the day there, they would like to have it included in the timestamp.

That was actually on purpose. The I'm fine with including the "hour:minute". But the tests need than to be adjusted too. This change seem to break the CI.

dalanicolai:
Could you specify which tests should get adjusted and how this breaks the CI? Of course this only adds the time to the time-stamp, which would be a great option to have for using it with [org-wild-notifier](https://github.com/akhramov/org-wild-notifier.el)(besides that it would be nice to have the option to schedule for a specific time instead of for a specific day only). If it breaks something we could just make it an option via an extra argument (and then make a separate command that calls this function with that extra argument, just like how the extra command for inserting at point was created in #335).

cslux:
https://github.com/bastibe/org-journal/blob/919abc914902d78ae7556c513e671db4a327e310/tests/org-journal-test.el#L293 This is the test which fails.

cslux:
I have fixed the test and merge this PR.

dalanicolai:
Haha! I just wanted to push a fix also. But anyway great! Thanks for merging.

Let me know if you have some questions about the other PR's, or if you would like me to add some extra comments to save you time from figuring out their purpose.

PR #344, opened 2021-04-18
Add a more robust set for `org-capture-templates` to README.org.

### LINKED ISSUE #343: Replace setq with `(setf (alist-get` for capture templates in the README.org.
**Describe the bug**
The default example suggests "seqting" org-capture-templates, which would overwrite it. I suggest replacing it with `setf `, which would add the template if it is not there, replace the entry if it as in:

That is, I have the following configuration snippet in my `init.el.`

```
(cl-labels
 ((org-journal-find-location ()
                             ;; Open today's journal, but specify a non-nil prefix argument in order to
                             ;; inhibit inserting the heading; org-capture will insert the heading.
                             (org-journal-new-entry t)
                             (unless (eq org-journal-file-type 'daily)
                               (org-narrow-to-subtree))
                             (goto-char (point-max))))
 (use-package org-journal
              :ensure t
              :demand t
              :after org
              :init
              ;; Change default prefix key; needs to be set before loading org-journal
              (setq org-journal-prefix-key "C-c j ")
              :config
              (setq org-journal-dir "~/Personal_Planner/Journal/"
                 org-journal-file-type 'monthly
                 org-journal-date-format "%F %A, %d %B %Y"
                 org-journal-time-format "%T %Z(%:::z)"
                 org-journal-enable-encryption nil)
              (setf (alist-get "j" org-capture-templates nil nil #'string-equal)
                 `("Journal entry"
                       plain
                       (function ,#'org-journal-find-location)
                       "** %(format-time-string org-journal-time-format)%^{Title}\n%i%?"
                       :jump-to-captured t :immediate-finish t)))
)
```

`cl-labels` hides the function from general visibility, and `setf` is a "generalised set", which would be clever enough to replace an association if it exists, and add one if it does not exist.

lockywolf:
- Replace a straightforward setq with a smarter setf/alist-get.

This should fix #343

### DIFF
--- a/README.org
+++ b/README.org
@@ -373,35 +373,49 @@ You can configure a capture template in order to integrate =org-journal= with =o
 as in the following example for a daily journal:
 
 #+BEGIN_EXAMPLE emacs-lisp
-  (defun org-journal-find-location ()
-    ;; Open today's journal, but specify a non-nil prefix argument in order to
-    ;; inhibit inserting the heading; org-capture will insert the heading.
-    (org-journal-new-entry t)
-    (unless (eq org-journal-file-type 'daily)
-      (org-narrow-to-subtree))
-    (goto-char (point-max)))
-
-  (setq org-capture-templates '(("j" "Journal entry" plain (function org-journal-find-location)
-                                 "** %(format-time-string org-journal-time-format)%^{Title}\n%i%?"
-                                 :jump-to-captured t :immediate-finish t)))
+(cl-labels
+ ((org-journal-find-location ()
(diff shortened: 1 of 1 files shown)


bastibe:
Thank you for your pull request. I see that in addition to the `(setf (alist-get`, you also changed the `defun` to `cl-labels`, and fit the code into `use-package`.

I would prefer it if you reverted these two changes, or presented them in a separate pull request for discussion. The reason being that `cl-labels` might not be as familiar to elisp novices as `defun`, and not everyone uses `use-package`. The README should focus on being easy to understand first and foremost, I think.

PR #345, opened 2021-04-25
Remove key definition in the package loading.

sstepashka:
Avoid modification of Emacs behavior on the package loading.

In order to follow [the coding convention](https://www.gnu.org/software/emacs/manual/html_node/elisp/Coding-Conventions.html), disable key definition in the package loading

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -423,8 +423,6 @@ This runs once per date, before `org-journal-after-entry-create-hook'.")
     (define-key calendar-mode-map (kbd "j s m") 'org-journal-search-calendar-month)
     (define-key calendar-mode-map (kbd "j s y") 'org-journal-search-calendar-year)))
 
-(global-set-key (kbd "C-c C-j") 'org-journal-new-entry)
-
 (defmacro org-journal--with-journal (file &rest body)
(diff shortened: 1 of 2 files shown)


cslux:
Hey @sstepashka! Good point. We should however also update the README.org file which mentions it in several places that `org-journal-new-entry` can be called with the key `C-c C-j`.

sstepashka:
Hey @cslux! I updated the document and added "C-c C-j" in `org-journal-mode`. It was mentioned in the documentation, but seems like it wasn't defined in there.

org-journal.el
```diff
@@ -399,7 +400,8 @@ This runs once per date, before `org-journal-after-entry-create-hook'.")
 
 ;; Key bindings
 (when (and (stringp org-journal-prefix-key) (not (string-empty-p org-journal-prefix-key)))
-  (let ((command-table '(("f" . org-journal-next-entry)
+  (let ((command-table '(("j" . org-journal-new-entry)
```
bastibe, line 403: This now contains "j" twice.

sstepashka: Fixed.

bastibe:
As this is a breaking change, it should probably get a (temporary) mention near the top of the README.

Otherwise, looks good to me!

PR #351, opened 2021-05-20
Document org-journal-hide-entries-p

mewfree:
I was looking how not to hide previous entries when running `org-journal-new-entry`, could not find anything in the README but found this variable while reading the source code.

I think it would be valuable to have it in the documentation for people like me who want to see their previous entries when writing a new one.


bastibe:
That's cool, thank you!

PR #359, opened 2021-07-07
Minor typos in docstrings

league:
I noticed a broken cross-reference to `org-journalf-file-type` and then went looking for any other dangling references in doc strings.


bastibe:
Thank you very much!

Issue #360, opened 2021-08-07
Multi-line file header is getting split and mixed with entry


Alexander-Miller:
**Describe the bug**
When `org-journal-file-header` is set to a value containing newlines only the first line stays at the top of the file, the rest becomes mixed up with the first entry.

**To Reproduce**

Configure org-journal like this:

```Elisp
(setf
 org-journal-file-type 'weekly
 org-journal-file-header "#+TITLE: Log %Y-%m-%d\n#+STARTUP: showall"
 org-journal-date-prefix "\n* ")
```

Call `org-journal-new-entry`.

The content of the journal buffer looks like this, the `#+STARTUP` header is inside the first entry:

```
#+TITLE: Log 2021-08-07

* Samstag, 07.08.2021
:PROPERTIES:
:CREATED:  20210807
:END:
#+STARTUP: showall
** 14:46 
```

**Expected behavior**
The header properties should all be at the top:

```
#+TITLE: Log 2021-08-07
#+STARTUP: showall

* Samstag, 07.08.2021
:PROPERTIES:
:CREATED:  20210807
:END:
** 14:46 
```

**Desktop (please complete the following information):**
 - OS: Manjaro
 - Emacs Version 28, self-compiled, using native compilation

PR #362, opened 2021-08-12
Fix multi-line file header being mixed up with the journal entry.

Alexander-Miller:
Potential fix for #360 - I've found out that ensuring the file header ends with a newline prevents it bleeding into the body of the entry.


cslux:
Approved 👍🏽 Thank very much for fixing it!

PR #371, opened 2021-12-22
Fixes 370

### LINKED ISSUE #370: `visual-line-mode` loaded by default but not present in `org-journal-mode-hook`
`visual-line-mode` is loaded by default for any `org-journal-mode` file, but the functions `visual-line-mode` or `turn-on-visual-line-mode` are not present in the variable `org-journal-mode-hook`. Instead, `org-journal-mode` calls `turn-on-visual-line-mode` explicitly [here](https://github.com/bastibe/org-journal/blob/71e8b10088ae52c4ac17f7af87020ea85fbc6ff7/org-journal.el#L389). Looking into the blame, this seems to be behaviour that was present since the [initial import](https://github.com/bastibe/org-journal/blob/22d3912f0d4dc05a1a47f52f8c696367cf6deeec/journal.el#L78) of `org-journal` into git.

Is this behaviour intentional?

quantumgc:
- Removed `turn-on-visual-line-mode` call
- Added `defcustom` definition for `org-journal-mode-hook`
- Added `visual-line-mode` to `org-journal-mode-hook`

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -386,7 +386,6 @@ This runs once per date, before `org-journal-after-entry-create-hook'.")
 (define-derived-mode org-journal-mode org-mode
   "Journal"
   "Mode for writing or viewing entries written in the journal."
-  (turn-on-visual-line-mode)
   (add-hook 'after-save-hook 'org-journal-after-save-hook nil t)
   (when (or org-journal-tag-alist org-journal-tag-persistent-alist)
     (org-journal--set-current-tag-alist))
@@ -1879,6 +1878,11 @@ Only one recipient is supported.  ")))
     (unless (equal org-journal-encrypt-on
                    'before-save-hook)
       (save-buffer))))
+
+(defcustom org-journal-mode-hook
+  '(turn-on-visual-line-mode)
+  "Hook to run when org-journal-mode is loaded."
+  :type 'hook)
 
 ;; Setup encryption by default
 ;;;###autoload
--- a/org-journal.el
+++ b/org-journal.el
@@ -1879,17 +1879,17 @@ Only one recipient is supported.  ")))
                    'before-save-hook)
       (save-buffer))))
 
+(defun org-journal-default-enable-encryption ()
+  "Add org-journal-encryption-hook to the hook org-journal-encrypt-on, enabling encryption by default."
+  (add-hook org-journal-encrypt-on
+            'org-journal-encryption-hook
+            nil t))
+
 (defcustom org-journal-mode-hook
-  '(turn-on-visual-line-mode)
+  '(turn-on-visual-line-mode
+    org-journal-default-enable-encryption)
   "Hook to run when org-journal-mode is loaded."
   :type 'hook)
-
-;; Setup encryption by default
-;;;###autoload
-(add-hook 'org-journal-mode-hook
-          (lambda () (add-hook org-journal-encrypt-on
-                               'org-journal-encryption-hook
-                               nil t)))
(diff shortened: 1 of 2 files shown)


quantumgc:
I've just realised that the default behaviour of `org-journal-mode-hook` should probably be mentioned in the documentation. I'll make that addition and update this PR.

I've also noted that the anonymous function
```elisp
(lambda () (add-hook org-journal-encrypt-on
                               'org-journal-encryption-hook
                               nil t))
```
is added to the hook directly afterwards using an `add-hook` call. I don't think that function can be removed using the `remove-hook` interface so should it be added to the `defcustom` call too?

org-journal.el
```diff
@@ -1880,6 +1879,11 @@ Only one recipient is supported.  ")))
                    'before-save-hook)
       (save-buffer))))
 
+(defcustom org-journal-mode-hook
+  '(visual-line-mode)
```
bastibe, line 1883: This toggles visual-line-mode instead of turning it on. I wonder if this might lead to problems if, say, org-mode-hook contained visual-line-mode as well?

bastibe:
> I've also noted that the anonymous function
> 
> ```emacs-lisp
> (lambda () (add-hook org-journal-encrypt-on
>                                'org-journal-encryption-hook
>                                nil t))
> ```
> 
> is added to the hook directly afterwards using an `add-hook` call. I don't think that function can be removed using the `remove-hook` interface so should it be added to the `defcustom` call too?

Good idea!

org-journal.el
quantumgc: Ah, thanks for catching that. I'll test it and change it if necessary.

quantumgc: It doesn't seem to have an effect, but I've made the change anyway.

org-journal.el
```diff
@@ -1880,12 +1879,17 @@ Only one recipient is supported.  ")))
                    'before-save-hook)
       (save-buffer))))
 
-;; Setup encryption by default
-;;;###autoload
-(add-hook 'org-journal-mode-hook
-          (lambda () (add-hook org-journal-encrypt-on
-                               'org-journal-encryption-hook
-                               nil t)))
+(defun org-journal-add-encryption-hook ()
```
bastibe, line 1882: I think the function name should not end on `-hook`, so as to not confuse people with the hook variable itself.

bastibe (changes requested):
Brilliant! Thank you!

I would only ask you to rename `org-journal-add-encryption-hook`. Something like `org-journal-enable-encryption` should to the trick.

(come to think of it, `org-journal-encryption-hook` has the same problem. But I don't know if we can change that without breaking existing functionality.)

quantumgc:
> (come to think of it, org-journal-encryption-hook has the same problem. But I don't know if we can change that without breaking existing functionality.)

I feel like there's also some confusion with `org-journal-encrypt-on` pointing to an actual hook so that might need to be renamed too (though at that point it feels like it needs a a separate issue/PR).

org-journal.el
quantumgc: Changed the function name to `org-journal-default-enable-encryption`, but I think the naming of the encryption functions should be rethought a little more in a separate issue/PR.

bastibe (approved):

org-journal.el
bastibe: I agree! But as you say, I'll leave that for another issue or pull request.

bastibe:
Thank you very much!

PR #374, opened 2022-01-02
Add copyright and license header

skangas:
Hi!

I am investigating if we could add this package to [NonGNU ELPA](http://elpa.nongnu.org/) (see link), a new Emacs Lisp package archive that will be enabled by default in Emacs 28. This means that users of that version or later will be able to install this package without any configuration: they can just run `M-x list-packages` and install it out of the box. We hope that this will improve Emacs and help bring new users to this package and others.

However, NonGNU ELPA has a requirement that all files to have a clear statement of its license and copyright status. I believe this patch would be sufficient to fulfill this requirement.

The rules for accepting a package into NonGNU ELPA are here, for your reference:

https://git.savannah.gnu.org/cgit/emacs/nongnu.git/tree/README.org#n133

The main difference between NonGNU ELPA and MELPA is that only stable versions of packages are released. A new release will be made automatically when you bump the ";;; Version: NNN" commentary header in the main file of this repository. This is similar to MELPA Stable, but we do not care about the git tag; only the commentary header.  In the future, you can bump the package version (thereby releasing a new version) when you think it makes sense and at your convenience.

Please let me know if you have any questions about NonGNU ELPA.

Thanks!

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -1,11 +1,39 @@
 ;;; org-journal.el --- a simple org-mode based journaling mode -*- lexical-binding: t; -*-
 
+;; Copyright 2013-2022 Bastian Bechtold
+
 ;; Author: Bastian Bechtold
 ;;         Christian Schwarzgruber
 
 ;; URL: http://github.com/bastibe/org-journal
 ;; Version: 2.1.2
 ;; Package-Requires: ((emacs "25.1") (org "9.1"))
+
+;; Redistribution and use in source and binary forms, with or without
+;; modification, are permitted provided that the following conditions are met:
+;;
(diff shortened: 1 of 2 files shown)


bastibe:
Thank you very much! I would love to have org-journal on NonGNU ELPA. From the documentation, it seems to me that I will have to build and upload every update to org-journal manually. Is this correct, or is there some automated machinery in place, as in MELPA stable?

skangas:
Great, thanks!

To answer your question, once this is setup a new package will automatically be built when you bump the ";; Version: NN" header. No need for manual intervention.

bastibe:
Thank you for the explanation

PR #376, opened 2022-01-05
Add NonGNU ELPA badge to README.org

skangas:
This package has now been added to NonGNU ELPA!

This commit adds a NonGNU ELPA badge as it looks nice, and is occasionally useful.

Thanks!


bastibe:
Thank you!

PR #377, opened 2022-01-05
Add NonGNU ELPA to installation instructions

skangas:
This updates the installation instructions to replace Marmalade (which is sadly down; maybe discontinued?) with NonGNU ELPA in the installation instructions.

It also fixes broken links to MELPA and MELPA Stable.

Thanks!


bastibe:
Again, thank you very much!

Issue #375, opened 2022-01-04
journal files are decrypted whenever the calendar is invoked.


dppdppd:
Every time I bring up the calendar to update a deadline or schedule in the agenda, org-journal decrypts all of my journal files.

I have org-journal-enable-agenda-integration set to nil.
I do not have journal files in my agenda-files.
I have confirmed that disabling org-journal eliminates this behavior.

Is this a bug? A misconfiguration on my part? 

my configuration:
>(use-package! org-journal
  :defer t
  :custom
  (org-journal-dir (concat org-root "/journal/"))
  (org-journal-file-format "%Y-%m.org")
  (org-journal-file-type 'monthly)
  (org-journal-date-format "%Y-%m-%d")
  (org-journal-created-property-timestamp-format "[%Y-%m-%d %a]")
  (org-journal-encrypt-journal t)
  (org-journal-enable-agenda-integration nil)
  :init
  )

bastibe:
I think that's necessary in monthly mode. How else could the calendar find out about the dates of your journal entries? The calendar integration is currently not optional.

That said, decryption is not strictly necessary in daily mode, but I don't know if the code is aware of that.

dppdppd:
It’s a pretty rough experience: each task (entirely unrelated to the journal) I update is a 10-20 second pause.

I find it hard to believe that anyone suffers through this so my setup must be unusual in one way or another.

bastibe:
You are indeed not alone: enabling `org-journal-enable-cache` should speed up calendar operations considerably.

Sorry for not mentioning that earlier. Since I don't use monthly/weekly files, nor encryption, I didn't keep up with all the cool tech @cslux and others built for these code paths.

dppdppd:
hmm. Well that certainly seems like the variable I want but it doesn't make a difference. All journal files are decrypted on every activation of the calendar.

When agenda integration is off and the calendar does not show journal entries, what is the purpose of parsing the journal contents?

dppdppd:
OK. So I removed the calendar hooks. Org-journal now doesn't bother me at all outside of org-journal usage. I have yet to noticed any ill effects yet, although I also only use a small subset of org-journal features.

Might be interesting to condition the hooks on agenda integration or a new calendar integration variable.

```
(use-package! org-journal
  :after org
  :custom
  (org-journal-dir (concat org-root "/journal/"))
  (org-journal-file-format "%Y-%m.org")
  (org-journal-file-type 'monthly)
[...]

PR #378, opened 2022-01-20
made calendar integration optional

dppdppd:
something like this?

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -271,6 +271,14 @@ It can be set to other hooks like `kill-buffer-hook'."
   "Add current and future org-journal files to `org-agenda-files' when non-nil."
   :type 'boolean)
 
+(defcustom org-journal-enable-calendar-integration t
+  "When non-nil, journal dates will be marked on the calendar.
+
+Additionally, calendar-specific commands will be keyboard-mapped.
+Users that don't want their journals decrypted when bringing up the calendar
+may consider setting this to nil. Also, see `org-journal-enable-cache'"
+  :type 'boolean)
+
 (defcustom org-journal-find-file 'find-file-other-window
   "The function to use when opening an entry.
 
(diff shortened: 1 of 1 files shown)


org-journal.el
```diff
@@ -405,9 +413,10 @@ This runs once per date, before `org-journal-after-entry-create-hook'.")
 
 
 ;;;###autoload
-(add-hook 'calendar-today-visible-hook 'org-journal-mark-entries)
-;;;###autoload
-(add-hook 'calendar-today-invisible-hook 'org-journal-mark-entries)
+(if org-journal-enable-calendar-integration
```
bastibe, line 416: I think this breaks the autoloads, which are necessary for making org-journal work in the calendar if it hasn't been invoked yet. So the branch probably needs to happen inside `org-journal-mark-entries` instead.

org-journal.el
```diff
@@ -434,21 +443,22 @@ This runs once per date, before `org-journal-after-entry-create-hook'.")
...
-    (define-key calendar-mode-map "[" 'org-journal-previous-entry)
-    (define-key calendar-mode-map (kbd "j n") 'org-journal-new-date-entry)
-    (define-key calendar-mode-map (kbd "j s f") 'org-journal-search-forever)
-    (define-key calendar-mode-map (kbd "j s F") 'org-journal-search-future)
-    (define-key calendar-mode-map (kbd "j s w") 'org-journal-search-calendar-week)
-    (define-key calendar-mode-map (kbd "j s m") 'org-journal-search-calendar-month)
-    (define-key calendar-mode-map (kbd "j s y") 'org-journal-search-calendar-year)))
+             do (define-key org-journal-mode-map (funcall key-func org-journal-prefix-key key) command))))
+
+(if org-journal-enable-calendar-integration
```
bastibe, line 448: Would it break anything if these remained enabled? Or perhaps it would make sense to only disable the first five. I could see a use case for `org-journal-new-date-entry` and the `org-journal-search-*` functions even with the calendar highlights disabled.

bastibe:
Something like this will work. But I added two comments that we should discuss or fix.

xeruf (approved):
have not manually tested, but looks good

casch-at:
Fixes #375

org-journal.el
casch-at: Good point.

casch-at:
@dppdppd Is it ok if I push to this branch, and do a squash merge afterwards?

PR #379, opened 2022-01-24
Skip marking calendar entries when calendar is just for input #375

dppdppd:
The more I thought about the previous PR, the less I liked it.

For one, it had the issues you raised, which is that it's not clear what calendar integration means wrt bindings. Should mark-entries still be mapped so users can manually mark them? And if they do, should the previously disabled commands be now enabled? And if none of the bindings are disabled, some of them would lead to confusion if the calendar isn't marked.

And secondly, regardless of what one does with the bindings, it didn't prevent the problem. An affected user would stall on calendar popups without any explanation or hint as to what package was doing it or how to fix it.

What do you think about this alternative? 
The calendar is not marked with journal entries when the calendar is popped up as a result of org-read-date. In my limited usage, journal entry dates aren't useful when picking a date for timestamp entry. But maybe they are in some contexts? More importantly, this eliminates the need to debug and identify the solution in the first place.

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -403,11 +403,23 @@ This runs once per date, before `org-journal-after-entry-create-hook'.")
 
 (defvar org-journal--search-buffer "*Org-journal search*")
 
-
+(defvar org-journal-dont-mark-entries nil)
+
 ;;;###autoload
 (add-hook 'calendar-today-visible-hook 'org-journal-mark-entries)
 ;;;###autoload
 (add-hook 'calendar-today-invisible-hook 'org-journal-mark-entries)
+
+(defun org-journal--org-read-date (orig-fun &rest args)
+  "In the case where the calendar is opened to input a timestamp, don't mark entries."
+  (progn
+    (setq org-journal-dont-mark-entries t)
+    (unwind-protect
+        (apply orig-fun args)
+      (setq org-journal-dont-mark-entries nil))))
+
+;;;###autoload
+(advice-add 'org-read-date :around #'org-journal--org-read-date)
 
 ;; Journal mode definition
 ;;;###autoload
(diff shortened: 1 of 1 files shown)


org-journal.el
```diff
@@ -1330,14 +1342,15 @@ from oldest to newest."
...
   (interactive)
-  (when (file-exists-p org-journal-dir)
-    (let ((current-time (current-time)))
-      (dolist (journal-entry (org-journal--list-dates))
-        (if (calendar-date-is-visible-p journal-entry)
-            (if (time-less-p (org-journal--calendar-date->time journal-entry)
-                             current-time)
-                (calendar-mark-visible-date journal-entry 'org-journal-calendar-entry-face)
-              (calendar-mark-visible-date journal-entry 'org-journal-calendar-scheduled-face)))))))
+  (unless org-journal-dont-mark-entries
```
bastibe, line 1345: This looks reasonable.

org-journal.el
```diff
@@ -403,12 +403,24 @@ This runs once per date, before `org-journal-after-entry-create-hook'.")
 
 (defvar org-journal--search-buffer "*Org-journal search*")
 
-
+(defvar org-journal-dont-mark-entries nil)
```
bastibe, line 406: This should probably be a `defcustom`, so we can attach some documentation and a type, right?

org-journal.el
```diff
@@ -403,12 +403,24 @@ This runs once per date, before `org-journal-after-entry-create-hook'.")
...
-

+(defvar org-journal-dont-mark-entries nil)
+
 ;;;###autoload
 (add-hook 'calendar-today-visible-hook 'org-journal-mark-entries)
 ;;;###autoload
 (add-hook 'calendar-today-invisible-hook 'org-journal-mark-entries)
 
+(defun org-journal--org-read-date (orig-fun &rest args)
```
bastibe, line 413: This seems to permanently advice `org-read-date` to always see `org-journal-dont-mark-entries` as `t`. Shouldn't this be contingent on `org-journal-dont-mark-entries` actually being set. And in that case, why do we need the `advice` anyway?

I guess I'm saying I don't quite get what this is doing. Could you explain it in a bit more detail?

bastibe (commented):
I think I like this approach better overall, as it is more localized. I have attached a few comments that I'd like to discuss, however.

dppdppd:
Of course.

We only mark entries when org-journal-dont-mark-entries is nil.
In the case of the org-read-date (which calls calendar), we toggle it on beforehand and back off right after.

So if the user calls (calendar) then they see journal markings. And if they call, say, (org-time-stamp), then they don't.

I did not make it customizable but can.

bastibe:
But why do we need to set `org-journal-dont-mark-entries` before calling org-read-date? I thought `org-journal-mark-entries` doesn't do anything anyway if `org-journal-dont-mark-entries` is `t`.

Also, I think a positive variable name is better than a negative one, so `org-journal-should-mark-calendar` instead of `org-journal-dont-mark-entries`. And the variable name should clarify that we're talking about calendar entries.

PR #380, opened 2022-01-24
Skip mark entries

dppdppd:
(ugh. renaming the branch reset the PR. never again...)

I flipped the logic so it might be clearer now.

We always mark entries, unless we call org-read-date.

```
(defun org-journal--org-read-date (orig-fun &rest args)
  "In the case where the calendar is opened to input a timestamp, don't mark entries."
  (progn
    (setq org-journal-allow-mark-entries nil) <----- turn it off temporarily
    (unwind-protect
        (apply orig-fun args)               <------- (calendar) gets called in here, and therefore org-journal-mark-entries
      (setq org-journal-allow-mark-entries t)))) <---- turn it back on
```

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -403,11 +403,23 @@ This runs once per date, before `org-journal-after-entry-create-hook'.")
 
 (defvar org-journal--search-buffer "*Org-journal search*")
 
-
+(defvar org-journal-dont-mark-entries nil)
+
 ;;;###autoload
 (add-hook 'calendar-today-visible-hook 'org-journal-mark-entries)
 ;;;###autoload
 (add-hook 'calendar-today-invisible-hook 'org-journal-mark-entries)
+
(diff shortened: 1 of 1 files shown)


bastibe:
This looks reasonable. I still don't quite understand why we have to disable marking explicitly for org-read-date, since the function that does the marking is disabled anyway. But if you say it's necessary, I'll trust your assessment.

But since the variable is supposed to be configured by the user, it should be a `defcustom`, not a `defvar`.

xeruf:
@dppdppd can you make the suggested correction so we can merge?
Then I would squash this with the title "Skip marking calendar entries for date input"

superhans2:
can you accept this pull request? I am also getting same errors that are reported in https://github.com/bastibe/org-journal/issues/397 and it makes journal unusable

PR #387, opened 2022-04-07
fill docstrings to prevent byte-compile warnings

jmay:
emacs 28 is aggressive about byte-compiling packages. This clears all the warnings about long docstrings.


bastibe:
Great! Thank you!

Issue #383, opened 2022-02-02
CREATED property at top level of journal file can cause add-entry to fail


jmay:
**Describe the bug**
If there is a top-level properties drawer in the journal file (outside any headings), and that drawer contains a `CREATED` property, then many journaling actions will abort with "Before first headline at position ... in buffer ...".

This is because `org-journal--created-re` seeks forward for the first `CREATED` entry, then seeks back with `(org-back-to-heading)` which fails because it is at top of file.

Quick easy workaround: Don't put a `CREATED` property at the top level of an org-journal file, because it is meaningless, those properties are only useful in the context of specific journal entries.

But journaling shouldn't break just because there's a useless property there.

**To Reproduce**
Go to the top of any journal file. Insert a properties drawer like
```
:PROPERTIES:
:ID: something
:CREATED: something
:END:
```
Then invoke `(org-journal-new-entry)` or `(org-journal-after-save-hook)`.

**Expected behavior**
Saving an org-journal buffer should ignore any top-level properties drawer contents.

**Desktop (please complete the following information):**
 - OS: MacOS 11.6
 - Emacs Version 28.0.90

bastibe:
Thank you for the bug report.

Would you like to contribute a pull request to fix this?

pnathan:
I also observe this specific bug.

emacs 27.2

xeruf:
Probably related:

    user-error: Created property timestamp format "[%Y-%m-%d]" doesn’t match CREATED property value ([2022-08-04 Thu 15:56]) from entry at line: Line 208

I get this if a subheading has a CREATED property (in my case created by Orgzly) - but weirdly enough only when the headline is folded.

xeruf:
This not only causes add-entry to fail, it causes creation of org timestamps to fail, everywhere! Highly annoying!

bastibe:
Very annoying indeed. These formatting differences can arise when more than one Emacs/computer add data to an org file with more than one locale setting.

If you know of a good way to fix this, I'd appreciate a pull request.

PR #388, opened 2022-04-07
don't break journaling if CREATED exists at file level #383

jmay:
The existence of a CREATED entry in the file-level property drawer causes
=org-journal-new-entry= to fail. Ignore all file-level properties when
searching through the file for the correct timestamp to start a new entry.


bastibe:
Wonderful! Thank you!

Could you restrict the pull request to just the code changes and remove the whitespace changes, please?

jmay:
Sorry, lost this in my email. Didn't mean to do any reformatting. New PR coming.

Issue #389, opened 2022-04-15
Add option to disable scheduled entries having a time


to-str:
When I am scheduling entries, I never use time. Basically my scheduling is based around days, but I don't work on the granularity of hours/minutes.

In org-journal, even if I don't enter any time (say for example I just put in `+2d` in org-agenda), it will add a timestamp showing `00:00`. And no matter how I look at it, that timestamp does not seem super useful to me.

What I get is this:

```
** TODO todo with time
<2022-04-17 Sun 00:00>
```

What I want is this:

```
** TODO todo without time
<2022-04-17 Sun>
```

I am very new to emacs (coming from neovim) and am not very smart at doing things the emacs way, but I played around a bit with the function for creating scheduled events and came up with this:

``` emacs-lisp
;; this needs to be set to either t or nil
(setq org-journal-schedule-include-time t)

(defun patch/org-journal-new-scheduled-entry (prefix &optional scheduled-time)
  "Create a new entry in the future with an active timestamp.
With non-nil prefix argument create a regular entry instead of a TODO entry."
  (interactive "P")
  (let ((time (or scheduled-time (org-time-string-to-time (org-read-date nil nil nil "Date:"))))
        org-journal-carryover-items)
    (when (time-less-p time (current-time))
      (user-error "Scheduled time needs to be in the future"))
    (org-journal-new-entry nil time)
    (unless prefix
      (insert "TODO "))
    (save-excursion
      (insert "\n")
      ;; only change is this line to make use of the variable
      (org-insert-time-stamp time org-journal-schedule-include-time))))
```

This works in my testing, I was wondering if there is any interest in this feature in general, something that could/would be merged maybe? I can ofc. create a PR if needed.

Best,
tostr

bastibe:
If you find it useful, I'm sure other people will find it useful, too! I'd be happy about a pull request that adds this as an option.

to-str:
PR is in #390. Closing this issue. Thanks @bastibe !

PR #390, opened 2022-04-24
Add feature to turn off scheduled entries having a timestamp.

to-str:
As per #389, I am adding a PR that adds this feature. First time working with this, and I am also on the move, so not quite sure everything is correct. But it _looks_ simple enough, hoping that it actually is.

Otherwise, please feel free to critique in any way, I should have a bit more time next week (but my schedule said I need to do this today, so I complied! ;))

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -244,6 +244,13 @@ When the current time is before the hour set by `org-extend-today-until'."
 By default, this is an org-mode sub-heading."
   :type 'string)
 
+(defcustom org-journal-schedule-include-time t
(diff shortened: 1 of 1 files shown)


to-str:
Ok, now I had a bit of time to figure out how to test. All three obvious cases work with expected behaviour (not set, set to `t` and set to `nil`).

bastibe:
Thank you very much, and sorry for forgetting this pull request for so long!

Would you like to add a note to the README for the new customization variable? Or should this remain in the customization menu only? I'm fine with either approach.

I'll merge the PR when you say it's ready.

PR #399, opened 2022-09-15
don't break journaling if CREATED exists at file level #383

jmay:
The existence of a CREATED entry in the file-level property drawer causes =org-journal-new-entry= to fail. Ignore all file-level properties when searching through the file for the correct timestamp to start a new entry.

(resubmit for PR388)


bastibe:
Wonderful, thank you!

Issue #404, opened 2022-12-01
ert test fails for org-journal


jmay:
I'm learning how `ert` works. `org-journal` has an `ert` test suite.

When I run the test suite, there is a single failure out of 10 tests.

I have a fix for this, but wanted to verify that others see this error, since I'm unfamiliar with the tool.

To run the tests, I am executing this from the command line in the org-journal directory:

`emacs -batch -l ert -l org-journal.el -l tests/org-journal-test.el -f ert-run-tests-batch-and-exit`

Full output from the batch command:

```
Running 10 tests (2022-11-30 16:44:19-0800, selector ‘t’)
   passed   1/10  org-journal-calendar-date-from-file-test (0.001045 sec)
   passed   2/10  org-journal-carryover-delete-empty-journal-test (0.830385 sec)
/private/var/folders/1w/9g63cjy91lx_yz74py0qgpw00000gn/T/org-journal-6kGuUO/20181231 and /var/folders/1w/9g63cjy91lx_yz74py0qgpw00000gn/T/org-journal-6kGuUO/20181231 are the same file
/private/var/folders/1w/9g63cjy91lx_yz74py0qgpw00000gn/T/org-journal-6kGuUO/20181231 and /var/folders/1w/9g63cjy91lx_yz74py0qgpw00000gn/T/org-journal-6kGuUO/20181231 are the same file
   passed   3/10  org-journal-carryover-items-test (0.045122 sec)
   passed   4/10  org-journal-carryover-keep-parents-test (0.014721 sec)
   passed   5/10  org-journal-convert-time-to-file-type-time-test (0.000429 sec)
   passed   6/10  org-journal-insert-header-test (0.023784 sec)
   passed   7/10  org-journal-scheduled-carryover-daily-test (0.015132 sec)
   passed   8/10  org-journal-scheduled-carryover-yearly-test (0.028051 sec)
Test org-journal-scheduled-weekly-test backtrace:
  signal(ert-test-failed (((should (equal (buffer-substring-no-propert
  ert-fail(((should (equal (buffer-substring-no-properties (point-min)
  (if (unwind-protect (setq value-104 (apply fn-102 args-103)) (setq f
  (let (form-description-106) (if (unwind-protect (setq value-104 (app
  (let ((value-104 'ert-form-evaluation-aborted-105)) (let (form-descr
  (let* ((fn-102 #'equal) (args-103 (condition-case err (let ((signal-
[...]

PR #405, opened 2022-12-01
fix failing test case #404

jmay:
Removed leading spaces from property-drawer lines.


bastibe:
Looks good to me! Thank you!

Issue #397, opened 2022-09-03
 Ignore unrecognized timestamps and timestamps in subheadings


xeruf:
### Description

I work with Orgzly, and this produces timestamps like:
```org
:PROPERTIES:
:CREATED: [2022-08-23 Tue 10:09]
:END:
```
While org-journal for me uses:
```
org-journal-created-property-timestamp-format

Value
"[%Y-%m-%d]"
```

#### Expected Results

org-journal ignores:
1. CREATED entries on subheadings
2. CREATED entries outside of PROPERTIES drawers (yes I moved it out to test and org-journal did not care)
3. CREATED entries that do not match its format

No errors, no complaining, just ignore them as if they weren't there :)

#### Actual Results

    user-error: Created property timestamp format "[%Y-%m-%d]" doesn’t match CREATED property value (nil) from entry at line: Line 637

xeruf:
Now this is really weird, if I call either of `org-time-stamp(-inactive)` anywhere as soon as org-journal is loaded, I get an error:

    user-error: Created property timestamp format "[%Y-%m-%d]" doesn’t match CREATED property value ([2022-08-23 Tue 10:09]) from entry at line: Line 602

Seems like org-journal is hooked in somewhere it shouldn't be?

xeruf:
See also https://github.com/bastibe/org-journal/issues/383 and https://github.com/bastibe/org-journal/issues/391

xeruf:
okay, org-journal hooks into the calendar to mark days. I see @jmay @dpppdppd you have already created patches, could you assist on this issue as well?
After an hour of debugging I think I found the culprit:
```elisp
(defun org-journal--file->calendar-dates (file)
  "Return journal dates from FILE."
  (org-journal--with-journal
      file
    (let (dates)
      (save-excursion
        (goto-char (point-min))
        (while (re-search-forward org-journal--created-re nil t)
          (when (= (save-excursion (org-back-to-heading) (org-outline-level)) 1)
            (push (org-journal--entry-date->calendar-date) dates)))
        dates))))
```
my journal contains `#+startup: overview`, which means the headings are all folded and `org-outline-level` always reports 1. Inserting `outline-show-all` in there did not seem to cut it :/

jmay:
@xeruf Make sure you restart your emacs and have the latest org-journal changes. Are you still getting errors? I can run `org-time-stamp` or `org-time-stamp-inactive` without problems. The calendar-dates function above was definitely where the patch was applied.

[...]

PR #407, opened 2023-01-03
fix: do not error out on timestamps with extra components

xeruf:
When a timestamp contains items beyond the date,
org-journal used to error out.
Now it simply ignores them.

Has been in production use by me in the last months.

fix https://github.com/bastibe/org-journal/issues/397

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -505,7 +505,8 @@ before it will be deposed."
     initially (setq format (regexp-quote (replace-regexp-in-string "%F" "%Y-%m-%d" format t)))
     for (fmt . rx) in org-journal--format-rx-alist
     do (setq format (replace-regexp-in-string fmt rx format t))
-    finally return format))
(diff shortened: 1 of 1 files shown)


bastibe:
I don't think I fully grasp the regexp magic you did there. Could you give an example of what `"[]>]$"` and `"[A-z0-9: ]*\\&"` refer to in practice? (and probably add that as a comment in the code as well)

xeruf:
honestly, I forgot too, and don't quite understand it anymore :sweat_smile: let me test what happens without it and report back

xeruf:
ah, now I know: it adds matching for weekday and time to the timestamp regex, to fix https://github.com/bastibe/org-journal/issues/397

bastibe:
So it replaces any remaining matchers in the square brackets with a non-capturing wildcard match? That sounds reasonable.

PR #409, opened 2023-02-24
Support a symlink directory as org-journal-dir

tomoyukim:
I found `org-journal-is-journal` fails to check the pattern of journal files in the symlink directory. I'm adding this PR to support a symlink directory as `org-journal-dir`.

In my understanding, `org-journal--dir-and-file-format->pattern` returns the pattern by chasing the symlink since it's implemented using `file-truename` but `buffer-file-name` returns a file symlink path.


bastibe:
Thank you for the pull request!

PR #419, opened 2023-10-22
Fix bug that hid previous date entries.

kizifumoli:
This commit resolves a bug that occurred in the "monthly" journal file setup. When in a file containing previous date's entries, running 'org-journal-new-entry would hide any entry not in the current date, regardless of the value of 'org-journal-hide-entries-p.


bastibe:
Looks good, thank you very much!

Issue #308, opened 2020-10-30
how to get org-journal-new-scheduled-entry behave similarly to org-schedule (i.e. "SCHEDULED:" keyword + correct duration)?


Ziip-dev:
**Describe the bug**
When scheduling a future appointment I run `org-journal-new-scheduled-entry`, select the corresponding day in the calendar view, then type e.g. `14:00+2` (everything seems to be interpreted correctly as the overlay indicates - cf. screenshot 1).
But the entry that `org-journal` generates when I validate has no `SCHEDULED:` keyword (unlike `org-schedule`), and the time of the appointment has shifted strangely (cf. screenshot 2).

**To Reproduce**
- Call `org-journal-new-scheduled-entry`
- Select a future day in the calendar view
- Type the time of the appointment manually with its duration
- Press `RET` to generate the entry

I doubt something this big really comes from a real bug and will be reproducible, but as I am still a beginner with `org-journal` I hope I can get some help here.

**Expected behavior**
A new entry in `org-journal` with the `SCHEDULED:` keyword like with `org-schedule` and the correct time and duration (cf. screenshot 3).

**Screenshots**
screenshot 1
![screen1](https://user-images.githubusercontent.com/62331138/97717364-ff0ebc00-1ac4-11eb-88f3-03edb1d884b2.png)

screenshot 2
![screen2](https://user-images.githubusercontent.com/62331138/97717372-03d37000-1ac5-11eb-9dea-c856efa90787.png)

screenshot 3
![screen3](https://user-images.githubusercontent.com/62331138/97718982-d982b200-1ac6-11eb-8f44-3a991b2ca040.png)

**Desktop (please complete the following information):**
 - OS: Ubuntu 18.04
 - Emacs Version 27.1

**Your Emacs Configuration**
```
(use-package org-journal
  :custom
  (org-journal-dir "~/Documents/org/roam/dailies/")
  (org-journal-file-type 'daily)
  (org-journal-date-prefix "#+TITLE: ")
  (org-journal-file-format "%Y-%m-%d.org")
  (org-journal-date-format "%A, %d %B %Y")
  :config
  (setq org-journal-find-file #'find-file-other-window
        org-journal-enable-agenda-integration t))
```

cslux:
[...]

Issue #48, opened 2015-07-15
orgmode assumes scheduled/deadline time from headline


k0377:
I use org-journal mostly as my first point when I need to write down something out of my mind. So I don't need to care about first, I can always refile the stuff later. A lot of my entries I make over the day are TODO items:

```
* Monday, 13. Jul 2015

** 08:15 some thoughts...
** TODO 10:23 something i have todo
SCHEDULED: <2015-07-17 Fri>
```

Now all of them start with a time which is exactly what I want but when I take a look in my agenda, this time is used as the timestamp for scheduled. Same happens, if I would define a deadline before.

```
Friday 17 July 2015
JOURNAL:        10:23......... Scheduled: TODO something i have todo
```

I don't like this behaviour, because this is wrong. Just because I started a journal entry on 10:23 and scheduled the task some days later, doesn't mean this is the time I'm gonna work on this task. I schedule a lot of tasks this way, without an exact time, this is just how my workflow looks. I don't need always a time for my tasks. The only workaround so far is, to remove the timestamp from the beginning of the journal entry or define a timestamp for the scheduled/deadline date as well.

Idk, maybe this is the default behaviour how orgmode works or is there something wrong with org-journal? Thanks for help! :kissing_cat:

bastibe:
I believe that org-journal has nothing to do with this. Org-journal does not change how the file is interpreted. It merely writes a few headlines and defines a few keyboard shortcuts.

I always thought the `TODO` keyword has to be the first word on the headline for it to be marked as a to do item?

k0377:
I guess orgmode assumes the time in the headline as scheduled time. Did you never make a todo item in your journal files? Do you know how I can prevent this behaviour?

bastibe:
I  use timed items for things I did, and TODOs for things I am planning to do, so there is typically no overlap. No idea how to prevent this behavior, sorry. But I would suggest asking on the org-mode mailing list or emacs.stackexchange.com.

k0377:
Thanks to @Rasmus http://article.gmane.org/gmane.emacs.orgmode/99084

`(setq org-agenda-search-headline-for-time nil)`

PR #420, opened 2023-11-05
Improvements to org-journal-new-scheduled-entry and org-journal-reschedule-scheduled-entry

TPKato:
This commit improves the behavior of `org-journal-new-scheduled-entry` and `org-journal-reschedule-scheduled-entry`.
More specifically,
- It behaves more like `org-schedule` (with "SCHEDULED: ").
- Inserting a time stamp `<...>` is now more consistent. (no meaningless `00:00`, time range can also be specified.)
- The time stamp (not the current time) is also added to a heading (in an org-mode way: `TODO` + time, not time + `TODO`).

### Related issues/pull requests:

- https://github.com/bastibe/org-journal/issues/308 .
- https://github.com/bastibe/org-journal/issues/389 (and https://github.com/bastibe/org-journal/pull/390 (closed)).
- https://github.com/bastibe/org-journal/pull/338
- https://github.com/bastibe/org-journal/issues/48 (closed)

### Description:

- `org-journal-new-scheduled-entry` will insert followings if a time is not specified.

  `M-x org-journal-new-scheduled-entry`
  `+1[RET]` or `S-<right>[RET]`

  ```
  ** TODO _
  SCHEDULED: <2023-11-04 Sa>
  ```

  (`_` shows a cursor position)

- If you specify a time, `org-journal-new-scheduled-entry` inserts time in a heading **and** as a timestamp with "SCHEDULED: ".

  `M-x org-journal-new-scheduled-entry`
  `23:59[RET]`

  ```
  ** TODO 23:59 _			; Note that this is NOT the current time, but the scheduled time.
  SCHEDULED: <2023-11-03 Fr 23:59>
  ```

  You may think that the time stamp after "TODO" is unneeded/redandant, but it is useful when you use `org-todo-list` (`C-c a t`):

  ```
  20231104:   TODO 23:59 I must do this.
  ```

- You can also specify time range.

  `M-x org-journal-new-scheduled-entry`
  `22:00+1[RET]`

  ```
  ** TODO 22:00 _
  SCHEDULED: <2023-11-04 Sa 22:00-23:00>
  ```

- Use `C-u` to avoid inserting "TODO".

  `C-u M-x org-journal-new-scheduled-entry`
  `23:59[RET]`
  ```
  ** 23:59 _
  SCHEDULED: <2023-11-04 Sa 23:59>
  ```

Note that `org-journal-reschedule-scheduled-entry` now only changes a timestamp with "SCHEDULED: " (and the time in a heading). Bare timestamp `<...>` without "SCHEDULED: " remains untouched.

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -731,7 +731,7 @@ This allows the use of `org-journal-tag-alist' and
           (org-set-tags org-crypt-tag-matcher)))
       (run-hooks 'org-journal-after-header-create-hook))))
 
-(defun org-journal--insert-entry (time org-extend-today-until-active-p)
+(defun org-journal--insert-entry (time org-extend-today-until-active-p &optional no-timestamp)
   "Insert a new entry."
   (unless (eq (current-column) 0) (insert "\n"))
   (let* ((day-discrepancy (- (time-to-days (current-time)) (time-to-days time)))
@@ -747,11 +747,11 @@ This allows the use of `org-journal-tag-alist' and
                          (format-time-string org-journal-time-format)))
                       ;; “time” is on some other day, use blank timestamp
                       (t ""))))
-    (insert org-journal-time-prefix timestamp))
+    (insert org-journal-time-prefix (if no-timestamp "" timestamp)))
   (run-hooks 'org-journal-after-entry-create-hook))
 
 ;;;###autoload
-(defun org-journal-new-entry (prefix &optional time)
+(defun org-journal-new-entry (prefix &optional time no-timestamp)
   "Open today's journal file and start a new entry.
 
 With a PREFIX arg, open the today's file, create a heading if it
@@ -805,7 +805,7 @@ hook is run."
       (goto-char (point-max)))
 
     (when should-add-entry-p
-      (org-journal--insert-entry time org-extend-today-until-active-p))
+      (org-journal--insert-entry time org-extend-today-until-active-p no-timestamp))
 
     (if (and org-journal-hide-entries-p (org-journal--time-entry-level))
         (outline-hide-sublevels (org-journal--time-entry-level))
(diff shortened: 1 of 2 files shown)


bastibe:
From a quick glance, this looks reasonable. But I'm not using this feature (any more), so I can't comment on whether this breaks existing workflows or not.

TPKato:
@jmay
I tried to write the tests, but I'm not sure if I wrote them correctly because this is my first time writing tests with ERT.
Feel free to modify or completely rewrite them ;-)
(And the test for org-journal-reschedule-scheduled-entry is very minimal.)

TPKato:
If we want to, we could apply the following additional patch to my pull request and introduce a new variable to make it less *breaking* (though I personally don't think this is necessary). Even with the following patch, it is still breaking, but for example `* 20:00 TODO` → `* TODO 20:00`, or `<2023-11-03 Fr 00:00>` → `<2023-11-03 Fr>` (removing the unnecessary `00:00`) would be considered bug fixes.

With this patch, the string `"SCHEDULED:"` will be inserted only if the user has set the variable (*1). If we consider it as an enhancement, this patch might be reasonable.

(*1)
To make my PR less breaking, I've set the default to `""`, but I strongly believe  it's better to set it to `org-scheduled-string` for the compatibility with org-mode. In this case, we can write something like this in CHANGELOG or README:
> If you want to avoid inserting "SCHEDULED:" before the timestamp (like the old version of org-journal), you can set the value of org-journal-schdeuled-string to `""` in your setup:
>     (setq org-journal-schdeuled-string "")

```
--- a/org-journal.el
+++ b/org-journal.el
@@ -394,6 +394,10 @@ This prefix key is used for:
 - `org-journal-search' (key \"s\")"
   :type 'string)

+(defcustom org-journal-scheduled-string ""	; or org-scheduled-string
+  "String added before a time stamp for schedules."
+  :type 'string)
+
 (defvar org-journal-after-entry-create-hook nil
   "Hook called after journal entry creation.")

@@ -1125,7 +1129,7 @@ With non-nil prefix argument create a regular entry instead of a TODO entry."
         (insert (format-time-string org-journal-time-format time)))
     (save-excursion
       (insert "\n"
-	      org-scheduled-string " ") ; "SCHEDULED: "
+	      org-journal-scheduled-string " ")
       (org-insert-time-stamp
        time org-time-was-given nil nil nil (list org-end-time-was-given)))))

@@ -1158,11 +1162,13 @@ With non-nil prefix argument create a regular entry instead of a TODO entry."
                     (insert (format-time-string org-journal-time-format time))))))
         ;; update time of timestamp (<...>)
         (org-back-to-heading)
-        (if (re-search-forward (concat org-scheduled-string "[[:blank:]]*" org-ts-regexp) nil t)
+        (if (re-search-forward (concat (regexp-quote org-journal-scheduled-string)
+                                       "[[:blank:]]*" org-ts-regexp)
+                               nil t)
             (replace-match "")
           (org-end-of-subtree)
           (insert "\n"))
-        (insert org-scheduled-string " ")
+        (insert org-journal-scheduled-string " ")
         (org-insert-time-stamp time org-time-was-given nil nil nil (list org-end-time-was-given))
         (org-cut-subtree))
       (let (org-journal-carryover-items)
```

bastibe:
Compatibility with different date formats is an issue I run into frequently. Any fix for that would be highly appreciated.

jmay:
Combined this PR with @TPKato's patch and PR #422, submitting a separate PR with the results

Issue #412, opened 2023-08-11
org-journal test suite failing under emacs 29


jmay:
org-journal test suite does not run cleanly (emacs 29.1)

Before I investigate further and work up a fix, I'd like some verification that this isn't just me.

- installed latest org-journal with elpaca under emacs 29
- at terminal `cd .emacs.d/elpaca/repos/org-journal`
- `eldev test`

**Expected behavior**
All tests should pass.

**Actual behavior**
```
3 unexpected results:
   FAILED  org-journal-carryover-delete-empty-journal-test
   FAILED  org-journal-carryover-items-test
   FAILED  org-journal-scheduled-weekly-test
```

**Desktop (please complete the following information):**
 - OS: MacOS 13.4.1
 - Emacs 29.1 installed from https://github.com/railwaycat/homebrew-emacsmacport/releases

Miciah:
I'm seeing the same tests fail during an update using Guix.

<details>

```text
starting phase `check'
Running 10 tests (2023-11-20 22:31:34+0000, selector ‘t’)
   passed   1/10  org-journal-calendar-date-from-file-test (0.000712 sec)
Test org-journal-carryover-delete-empty-journal-test backtrace:
  signal(error ("Calling ‘org-fold-core-region’ with missing SPEC"))
  error("Calling `org-fold-core-region' with missing SPEC")
  org-fold-region(35 41 t drawer)
  org-insert-property-drawer()
  org-get-property-block(1 force)
  org-entry-put(nil "CREATED" "20190101")
  org-set-property("CREATED" "20190101")
  (progn (insert "* Wednesday, 01/01/19\n") (org-set-property "CREATED
  (unwind-protect (progn (insert "* Wednesday, 01/01/19\n") (org-set-p
  (save-current-buffer (set-buffer temp-buffer) (unwind-protect (progn
  (let ((temp-buffer (generate-new-buffer " *temp*" t))) (save-current
  (let ((buffer "20181231") (org-journal-carryover-delete-empty-journa
  (let* ((org-journal-dir (concat org-journal-dir-test "-link")) (comm
  (closure (t) nil (let* ((org-journal-dir (concat org-journal-dir-tes
  ert--run-test-internal(#s(ert--test-execution-info :test #s(ert-test
  ert-run-test(#s(ert-test :name org-journal-carryover-delete-empty-jo
  ert-run-or-rerun-test(#s(ert--stats :selector t :tests ... :test-map
[...]

PR #422, opened 2023-11-21
revise tests so they run cleanly on emacs 29.1 #412

jmay:


TPKato:
Using fset has many negative effects (in particular, running ert-run-tests-interactively is a disaster), so I think it is better to use another method. I don't know if it's the best, but cl-letf seems safer.
Applying the following patch (after your PR #422) will fix this problem and all tests should pass (tested on emacs version 29.1).

```
diff --git a/tests/org-journal-test.el b/tests/org-journal-test.el
index 4e5b6e2..0b3baa9 100644
--- a/tests/org-journal-test.el
+++ b/tests/org-journal-test.el
@@ -227,10 +227,11 @@
         (save-buffer)
         (kill-buffer)
         (let ((message-marker nil))
-          (fset 'message (lambda (x y) (setq message-marker x)))
-          (org-journal-read-or-display-entry (encode-time 0 0 0 2 1 2019) 'noselect)
-          (should (equal "No journal entry for this date." message-marker))
-          ))))
+          (cl-letf (((symbol-function 'message)
+                     #'(lambda (x &rest y) (setq message-marker x))))
+            (org-journal-read-or-display-entry (encode-time 0 0 0 2 1 2019) 'noselect)
+            (should (equal "No journal entry for this date." message-marker))
+            )))))
 
 (ert-deftest org-journal-search-build-file-list-test ()
   "Test for `org-journal--search-build-file-list'."
@@ -348,7 +349,7 @@
                ;; "   "
                ;; org-scheduled-string
                ;; " "
-               (scheduled-string (concat "<" (format-time-string (cdr org-time-stamp-formats) scheduled-entry-time) ">")))
+               (scheduled-string (concat "SCHEDULED: " "<" (format-time-string (car org-time-stamp-formats) scheduled-entry-time) ">")))
           ;; Add first scheduled entry
           (org-journal-new-scheduled-entry nil scheduled-entry-time)
           (insert "Task 1")
```

bastibe:
Thank you for working on this!

jmay:
@TPKato Thanks for this. I applied your patch, but only the first hunk. The second hunk was causing the scheduled-weekly-test to fail due to the extra "SCHEDULED:" text. Perhaps this is necessary for #420 ? I don't want to merge these changes until they both pass independently.

PR #424, opened 2023-12-06
Combination of 2 existing PRs plus patches: scheduling timestamps and clean tests

jmay:
I *think* that with this change, both of the other two PRs can be closed. @bastibe @TPKato


casch-at:
As it was broken for me too, at least until now :-) I will be happy to (review and) test it as well.

casch-at:
Looks good to me! Thank you @jmay and @TPKato.

PR #442, opened 2024-08-25
Add support for age encryption via age.el

### LINKED ISSUE #441: Support for age encryption
It would be great to support [age](https://github.com/FiloSottile/age) for encryption as well as the existing gpg support.

There is already [age.el](https://github.com/anticomputer/age.el) which might help.

It might be enough to even allow the setting of a custom file extension, if it is possible to say make the file extension `.org.age` and age.el is already installed I suspect it would transparently encrypte/decrypt.

fapdash:
- require age.el if the package is installed https://github.com/anticomputer/age.el
- add new customizable variable `org-journal-encryption-extension` to control the file extension appended to the file name to trigger transparent encryption
- recognize `.age` files as journal files
- generate encrypted files based on chosen file extension instead of hard coding `.gpg`
- support `.age` files in `org-journal-re-encrypt-journals` - only re-encrypt files for the currently chosen encryption extension - functional change: don't edit/re-save files that aren't encrypted. I assume that they where handled by this function in the past was an oversight and not intended. - allow multiple recipients for `age` encryption

closes #441

---

age encyrption works with this change.
TODOs from age encrypted journal files get added to the agenda. (`org-journal-enable-agenda-integration`)

I wasn't sure how to do optional dependencies for Emacs packages.
If someone tries to set `age` as extension the `org-journal-re-encrypt-journals`
function will fail because of missing functions, of course.

`org-journal-re-encrypt-journals` works for me if `age-default-recipient` and `age-default-identity` are set,
but things got a bit weird when I hadn't set those.
I might be missing some settings, but at some point age encryption wasn't working at all and I had to restart Emacs.

Theoretically org-journal would work with any encryption that implements transparent, as long as the `org-journal--list-files` and `org-journal--dir-and-file-format->pattern` would get consider the value of `org-journal-encryption-extension` in their regexes. If we don't want to implement this then it's probably better to make `org-journal-encryption-extension` of `:type '(choice (const :tag "age" age) (const :tag "gpg" gpg))` instead of `'string`.

### Instructions for testing

- Get `age` binaries: https://github.com/FiloSottile/age/releases/latest
- Place binaries in $PATH

Create age keys:

```bash
age-keygen -o key.txt
age-keygen -y key.txt > key.pub
```

Example `age.el` configuration for testing:

```elisp
(use-package age
  :ensure t
  :demand t
  :custom
  (age-program "age")
  (age-default-identity "~/.age/key.txt")
  (age-default-recipient '("~/.age/key.pub"))
  :config
  (age-file-enable))
```

Activate `age` encrption for `org-journal`:

```elisp
(setq 
  org-journal-encrypt-journal t
  org-journal-encryption-extension "age")
```

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -736,34 +736,34 @@ This allows the use of `org-journal-tag-alist' and
     (unless (if (org-journal--daily-p)
                 (or (search-forward entry-header nil t) (and (goto-char (point-max)) nil))
               (cl-loop
-                with date = (decode-time time)
-                with file-dates = (sort (org-journal--file->calendar-dates (buffer-file-name))
-                                        (lambda (a b)
-                                          (org-journal--calendar-date-compare b a)))
-                with entry
-                initially (setq date (list (nth 4 date) (nth 3 date) (nth 5 date)))
-                unless file-dates ;; New entry at bof
-                do
-                  (unless (re-search-forward (concat "^\\(" org-outline-regexp "\\)") nil t)
-                    (goto-char (point-max)))
-                  (if (org-at-heading-p)
-                      (progn
-                        (beginning-of-line)
-                        (insert "\n")
-                        (forward-line -1))
-                    (forward-line -1)
-                    (end-of-line))
-                and return nil
-                while file-dates
-                do
-                  (setq entry (car file-dates)
-                        file-dates (cdr file-dates))
-                if (or (org-journal--calendar-date-compare entry date) (equal entry date))
-                do
-                  (org-journal--search-forward-created entry)
-                  (when (org-journal--calendar-date-compare entry date) ;; New entry at eof, or somewhere in-between
-                    (org-end-of-subtree))
-                and return (equal entry date))) ;; If an entry exists don't create a header
+                    with date = (decode-time time)
+                    with file-dates = (sort (org-journal--file->calendar-dates (buffer-file-name))
+                                            (lambda (a b)
+                                              (org-journal--calendar-date-compare b a)))
+                    with entry
+                    initially (setq date (list (nth 4 date) (nth 3 date) (nth 5 date)))
+                    unless file-dates ;; New entry at bof
+                    do
+                    (unless (re-search-forward (concat "^\\(" org-outline-regexp "\\)") nil t)
+                      (goto-char (point-max)))
+                    (if (org-at-heading-p)
+                        (progn
+                          (beginning-of-line)
+                          (insert "\n")
+                          (forward-line -1))
+                      (forward-line -1)
+                      (end-of-line))
+                    and return nil
+                    while file-dates
+                    do
+                    (setq entry (car file-dates)
+                          file-dates (cdr file-dates))
+                    if (or (org-journal--calendar-date-compare entry date) (equal entry date))
+                    do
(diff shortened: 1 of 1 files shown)


fapdash:
@rosstimson Can you please test if this is working for you?

rosstimson:
@fapdash Thanks for your work here.  I don't think it is encrypting the file properly.

My relevant config looks like:

```elisp
(use-package age
    :mode ("\\.age\\'" . age-encryption-mode)
    :custom
    (age-default-identity (expand-file-name ".age/rosstimson.txt" (getenv "HOME")))
    ;; My public age key
    (age-default-recipient "age17ll5hultu5gpye40nlj7z7ktcxl8ne9xvtd6udzugkmwlmt83uns82smww")
    :config
    (age-file-enable))

(use-package org-journal
    :after org
    :bind ("C-c j" . org-journal-new-entry)
    :custom
    (org-journal-file-type 'monthly)
    (org-journal-date-format "%A, %d %B %Y")
    (org-journal-encrypt-journal t)
    (org-journal-encryption-extension "age")
    (org-journal-dir (concat (concat org-directory "/journal/") (format-time-string "%Y"))))
```

If I create a test file of `foo.age` and save some text then check it in terminal with `less` I can see that it is age encrypted and it automatically decrypts when I open it within emacs.

I then shifted my existing gpg encrypted journal files into a backup dir and used `org-journal-new-entry` to create a new journal entry, when I check the file in the terminal it shows as plain text rather than encrypted, it does indeed have have the `.age` extension though.

I've rebuilt the org-journal package with your branch and restarted emacs of course, if I enter the customize stuff I can see your newly added `org-journal-encryption-extension` when I search for it, I saw no errors with my init config.

fapdash:
@rosstimson Mh, I can't reproduce this.. :shrug: 
`org-journal` doesn't really change anything about `age.el`, the only change you're hitting is that the file extension is changed to `.age`, which you've confirmed is working.

Can you please check if `age-encryption-mode` is active in the buffer for your journal `.age` file after you've created it via `org-journal-new-entry`?
I don't expect it to make a difference, but can you try without the `:mode ("\\.age\\'" . age-encryption-mode)` line? (mode activation is already handled by `age-file-enable`)

rosstimson:
@fapdash Not sure why it was a problem but you were correct, simply removing the line `:mode ("\\.age\\'" . age-encryption-mode)` has fixed it.  So exact same config as I posted above minus that link.

I can now confirm that this works as expected.  Thanks again for your work here.  🎉

rosstimson:
@bastibe Any chance of getting this merged?  Just been caught out after I cleared package manager caches and realised this broke as I'd pulled these changes to test.  I've now had to switch to the fork/branch.

bastibe:
Is this working? I can't test it at the moment, but if you and @fapdash say it's fine, I'll happily merge it.

rosstimson:
Yeah, this works fine for me with config I posted.

fapdash:
@bastibe I'm not 100% on the `org-journal-re-encrypt-journals`, but I vaguely remember doing manual testing and it worked.
I've been using age encryption with these changes since September last year, so that part definitely works.

bastibe:
Thank you very much!

Issue #451, opened 2025-01-02
First Week Of Year file-format - %Y-%V - shows correct week no (1), but wrong year (2024) for 2.1.2025


triplem:
**Describe the bug**
If a new file is created in the first week of a new year (well, at least for 2025), the file-fomat shows the correct week-no, which is no. 1, but still uses the "old" year, meaning the year on the beginning of the week.

**To Reproduce**
Create a weekly journal with the following settings:

  (setq org-journal-file-type 'weekly)
  (setq org-journal-start-on-weekday calendar-week-start-day)
  (setq org-journal-file-format "%Y-%V.org")

This will create a new file with the name 2024-01.org on 2.1.2025.

**Expected behavior**
The first week of the year should be in the correct year, meaning 2025 in this case.

drewbenn:
Short answer: use `%G` instead of `%Y`, but also update `(org-journal--format-rx-alist)` to support `%G`. Though now that you've made it through the beginning of 2026, this shouldn't be a problem again until [2030](https://www.epochconverter.com/weeks/2030)!

---

So as you know, org-journal picks a filename based on the day at the beginning of the week. When it moved backward to Monday, December 30 and queried that day's year and that day's ISO week, there was a mismatch. If it had queried that day's _ISO year_, on the other hand, the result would have been in sync with the ISO week.

- Change your file format setting to "%G-%V" so it will use the ISO week for both year and week.

But if you just use `%G` then a different lookup will fail while trying to create the CREATED property (giving you an error like, `org-fold-region: Calling 'org-fold-core-region' with missing SPEC`). That's why there is a comment in the code listing only a few supported date fields, even though the emacs function being used ([`(format-time-string)`](https://www.gnu.org/software/emacs/manual/html_node/elisp/Time-Parsing.html#index-format_002dtime_002dstring)) supports many more. Updating the lookup table to add support for `%G` will create the new file properly.

- In `(org-journal--format-rx-alist)` change the `"%Y"` to `"%[YG]"`

I tested all this by picking a date in January where the ISO week started (Monday) in December: 2020-01-01. Then created an entry for that date, using each of 3 different `org-journal-file-format` values:

> (setq org-journal-file-format "%Y%m%d")  ;default
(setq org-journal-file-format "%Y-%V")   ;per bug
(setq org-journal-file-format "%G-%V")   ;what I think it should be
>
>then use: 
>- M-x org-journal-new-date-entry Enter [2020-01-01]
>
>and see what filename you get:
[...]

PR #462, opened 2026-04-02
Add support for %G (ISO year)

drewbenn:
Support %G (ISO year) in filenames, and update comments to match.
Allows users to name files correctly with ISO weeks when week 1 crosses a year boundary (see #451).

### DIFF
--- a/org-journal.el
+++ b/org-journal.el
@@ -175,16 +175,21 @@ currently to be done manually by calling `org-journal-invalidate-cache'."
 (defcustom org-journal-file-format "%Y%m%d"
   "Format string for journal file names (Default \"YYYYMMDD\").
 
-This pattern MUST include `%Y', `%m' and `%d' when `org-journal-file-type' is
-`daily' or `weekly'. When `org-journal-file-type' is `monthly' this pattern
-MUST at least include `%Y' and `%m', and at least `%Y' when
-`org-journal-file-type' is `yearly'.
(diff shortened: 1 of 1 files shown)


org-journal.el
```diff
@@ -175,16 +175,21 @@ currently to be done manually by calling `org-journal-invalidate-cache'."
 (defcustom org-journal-file-format "%Y%m%d"
   "Format string for journal file names (Default \"YYYYMMDD\").
 
-This pattern MUST include `%Y', `%m' and `%d' when `org-journal-file-type' is
-`daily' or `weekly'. When `org-journal-file-type' is `monthly' this pattern
-MUST at least include `%Y' and `%m', and at least `%Y' when
-`org-journal-file-type' is `yearly'.
+This pattern MUST include enough format specifiers to uniquely identify
+the filename for the date and file type:
+  | org-journal-file-type | pattern MUST include | or MUST include |
```
bastibe, line 180: if I remember correctly, docstrings are not formatted, so tables are probably not allowed.

drewbenn: looks okay to me with `C-h v` (though I'm using emacs 30.1 if that matters) but I'm happy to change it if you'd prefer?

Image

I'm not entirely happy with the table but thought it was getting to the point where there were too many details for a sentence.

bastibe:
Thank you!

Issue #439, opened 2024-07-20
Question: How to fold previous days in a weekly journal?


anthonyclarka2:
**Describe the bug**
When I call `org-journal-new-entry` all the previous days in a weekly journal are expanded. 

Question: Is there a way to fold previous days?

**This almost certainly isn't a bug in org-journal** I hope I am not causing upset by asking this here :/

**Screenshots**
This is what my current week journal looks like before I start:
![org-journal-before-new-entry](https://github.com/user-attachments/assets/be94340c-7cc9-45d4-bb80-e5e163facbf8)

After running `org-journal-new-entry`:
![org-journal-after-new-entry](https://github.com/user-attachments/assets/795daed3-0f91-4b72-b4aa-66ae655cdb00)

How I want it to look:
![org-journal-desired-behaviour](https://github.com/user-attachments/assets/800378f9-c866-4464-aec9-b871391045af)

**Desktop (please complete the following information):**
 - OS: Fedora 40, x86_64, Wayland, Nvidia 4070
 - Emacs Version 31.0.50

**Your Emacs Configuration**
```lisp
;; https://github.com/bastibe/org-journal
(use-package org-journal
  :bind
  ("C-c n j" . org-journal-new-entry)
  ("C-c j" . org-journal-open-current-journal-file)
  :custom
  ;; (org-journal-date-prefix "#+TITLE: ")
  ;; (org-journal-time-prefix "* ")
  (org-journal-file-format "%Y-%m-%d.org")
  (org-journal-file-type 'weekly)
  (org-journal-dir "~/Org/Journal/")
  (org-journal-date-format "%Y-%m-%d")
  (org-journal-carryover-items "TODO=\"TODO\"|TODO=\"NEXT\"|TODO=\"PROG\"")
  (org-journal-enable-agenda-integration t)
  (org-journal-find-file 'find-file))
```

bastibe:
Does the org variable `org-set-startup-visibility` change this behavior?

aclarknexient:
(work account)

Unfortunately not, it didn't change the behaviour.

bastibe:
There's also `org-journal-hide-entries-p` that you could customize

aclarknexient:
I will look into that, thank you very much. I won't be able to get to it until the weekend, is it OK if this issue remains open until then?

bastibe:
No problem whatsoever. And no rush, either.

aclarknexient:
Here's what I've discovered:

[...]

PR #463, opened 2026-04-12
Option to hide titles of previous days' entries when creating a new entry #439

drewbenn:
Enhancement for #439: changes `org-journal-hide-entries-p` to allow a new value, `'previous` which provides the requested behavior: hiding all journal entries in previous days.


bastibe:
I like this! But as I don't use Emacs any longer, I'd love to hear another contributor's opinion about it. If you can find another person to have a look, I'll happily merge it.

Ping me again if you can't find anybody.

PR #464, opened 2026-09-20
342: Don't leave previous buffer narrowed when creating a new day's entry

drewbenn:
Applies to weekly/monthly/yearly journals only. When creating any entry on the first day, all the entries except the last in the previous week/month/year's buffer get folded. This prevents search from finding any of those folded entries.

Solve this by applying save-restriction to the correct buffer (the previous one) when folding it to get carryover items.

Also, since that buffer's view gets changed by us, reset its sublevel view to the org-journal default, if that is enabled.


drewbenn:
NB this is a simple re-submit of #460 after some local branch management forced that issue closed. Nothing new or particularly interesting here.

